PR #5, opened 2017-09-05
Add (broken) test for `get_definition()`

thisfred:
In a function like the following, parso fails to find the definition for
the second occurrence of `name`.

```python

    def fun():
        name = 1
        bar = name + 1

```

get_definition dutifully walks up the tree from `name + 1`, but bails when it gets to the `suite` (that I think represents the function body?) That means maybe the `suite` should return the definition in `get_defined_names()`, but I'm not sure that's enough, because if the lookup can't walk past this `suite` it could never find the function parameters as definitions, for instance...

### DIFF
--- a/test/test_parser_tree.py
+++ b/test/test_parser_tree.py
@@ -149,3 +149,19 @@ def test_returns():
 
     r, = get_return_stmts('def x(): return 1')
     assert r.type == 'return_stmt'
+
+
+def test_get_definition_finds_definition_in_same_function(each_version):
+    code = dedent('''
+        def fun():
+            name = 1
+            bar = name + 1''')
+    module = parse(code, version=each_version)
+
+    # the first occurrence of name is also where it is defined
+    definition = module.get_name_of_position((3, 8)).get_definition()
+
+    # `name` occurs twice, both occurrences should have the same definition
+    # (which is the first occurrence.
+    assert [definition, definition] == [
+        o.get_definition() for o in module.get_used_names()['name']]


thisfred:
Since there are no tests that I can find for `get_definition` maybe it isn't meant to be part of the API (just yet)?

davidhalter:
Hi @thisfred

First off, the tests don't exist, because I was too lazy to write them also for parso. It's pretty well tested in Jedi (It's used for all type inference there).

The issue you found is not an actual issue. It's just not documented well enough IMO. Type inference is not part of the scope of parso. Type inference is part of jedi. The function `get_definition` is basically just a helper for `is_definition` that might come in handy if you need that information. 

In your example what happens if there's an if and there are suddenly two definitions? It won't really work.

I'm open to suggestions about renaming this function and/or documenting it better. I can understand that it's a bit confusing.

thisfred:
Thanks for the explanation! It's a shame, as I was hoping to replace a large chunk of my refactoring library with parso, but for that to work, for instance for rename refactorings, it'd be really handy to have an easy way to find which names actually refer to the same thing, rather than just having the same name. I have working code for this based on Python's AST module, but it's a lot of code and there are a lot of hairy edge cases, so I was kind of hoping to be able to throw all of that complicated code away. ;)

davidhalter:
That's what Jedi's for. `jedi.Script.usages` is still kind of broken, but we're getting there. It's really only a matter of time until it is usable.

thisfred:
I gathered that after listening to https://www.podcastinit.com/episode-113-jedi-code-completion-with-david-halter/ I'll evaluate jedi itself as a potential dependency, and see if there's a way I can contribute there.

PR #8, opened 2017-10-01
Add support for python 3.7

### LINKED ISSUE #7: No grammar for Python 3.7
Using IPython 6.2 with Python 3.7.0a1 (early adopter so to say), I get an error when trying to tab-complete which refers to a missing grammar text file. Copying the grammar for 3.6 to 3.7 resolves the problem.

Could you deliver a preliminary grammar for Python 3.7 by using the grammar from Python 3.6?

haakenlid:
grammar37.txt is simply a copy of grammar36.txt, since there's no syntax
changes in 3.7

https://docs.python.org/3.7/whatsnew/3.7.html

fixes #7

### DIFF
--- a/parso/python/grammar37.txt
+++ b/parso/python/grammar37.txt
@@ -1,0 +1,150 @@
+# Grammar for Python
+
+# NOTE WELL: You should also follow all the steps listed at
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks!

PR #11, opened 2017-10-23
Fix the cache path

mmorearty:
Any caller of Parso can specify a cache path, and if none is specified, Parso will fall back to a default.

Parso's code to read a file from the cache handled this correctly; but its code to write a file to the cache had a bug -- any override of the default was ignored.

In the case of Jedi (and therefore jedi-vim), this is a significant problem, since Jedi overrides the default. The result is that files will be written to the cache, but will then never be found when reading from the cache.


davidhalter:
Wow! Thanks for finding and fixing this. I really didn't see it coming. We should probably add a test for this some time.

mmorearty:
So glad I could help!

PR #13, opened 2017-10-25
Minor documentation fixups

robodair:
I'm working on https://github.com/robodair/pydocstring to use parso instead of regex, and hoped to clear up a few bits and pieces of the parso docs while I'm at it.

### DIFF
--- a/parso/tree.py
+++ b/parso/tree.py
@@ -278,6 +278,14 @@ class BaseNode(NodeOrLeaf):
         return self._get_code_for_children(self.children, include_prefix)
 
     def get_leaf_for_position(self, position, include_prefixes=False):
+        """
+        Get the :py:class:`parso.tree.Leaf` at ``position``
+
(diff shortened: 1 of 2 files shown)


parso/tree.py
```diff
@@ -278,6 +278,14 @@ def get_code(self, include_prefix=True):
         return self._get_code_for_children(self.children, include_prefix)
 
     def get_leaf_for_position(self, position, include_prefixes=False):
+        """
+        Get the :py:class:`parso.tree.Leaf` at ``position``
+
+        :param tuple position: A position tuple, row, column. Rows start from 1
+        :param bool include_prefixes: If ``False``, ``None`` will be returned if ``position`` falls
+            on whitespace before a leaf
```
davidhalter: Can you maybe also add `if position falls on whitspace or comments before a leaf`?

davidhalter:
Thanks for the improevements. I hope you like it :)

PR #14, opened 2017-10-27
Add get_scope method to PythonMixin

robodair:
Helpful for navigating up the tree when I'm finding the scope to generate a docstring for!

### DIFF
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -110,6 +110,20 @@ class PythonMixin(object):
                     return result
         return None
 
+    def get_scope(self):
+        """
+        Get the first :py:class:`parso.python.tree.Scope` in the parents of this
+        :py:class:`parso.python.tree.PythonMixin`.
+
+        :return: the first :py:class:`parso.python.tree.Scope` above this in the tree
+        """
+        def get_enclosing_scope(node):
+            if isinstance(node.parent, Scope) or node.parent is None:
+                return node.parent
+            else:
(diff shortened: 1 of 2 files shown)


davidhalter:
I'll keep it open for now. I had a function in parso called `get_parent_scope` that did essentially the same. I'm not sure I want it here. It's now in https://github.com/davidhalter/jedi/blob/master/jedi/parser_utils.py.

I'll think about reintroducing it.

robodair:
Should I also add a check for being a flow to match that functionality?

davidhalter:
No it's fine for now. Not sure how I would want this function to look like at the moment. There's a reason why I removed it in the first place - scopes are a bit strangely defined.

robodair:
Would it be better to have a more flexible implementation that gets given a list of types to look for? Or should something like this just be implemented in user's code instead?

davidhalter:
Well, there's tree.search_ancestor if you need something like that. I don't think we need to have everything in parso. Methods should be in parso if they are usable for at least 50% of the projects and if they are clearly defined. Otherwise write the code in your code base. It's what I did in jedi as well. There's a ton of code that's not part of parso but actually uses it.

robodair:
Good enough for me!

davidhalter:
Didn't say I will never merge, just not now. I'll keep it open.

davidhalter:
I'm not looking at merging this anytime soon. There is a chance that I will add it in the future, but for now I will close.

PR #16, opened 2017-10-27
Iter raise statements in a Function

robodair:
This one isn't quite done yet, I'd like some input on whether this should look ahead and try to determine if a found exception is inside a try-catch or not.

(Because from my use-case for this there's no point in yielding an exception that gets caught)


davidhalter:
@robodair I think this certainly goes into the right direction. IMO we shouldn't check if it was raised in try/except. That is way to complicated, considering that there might be a raise again and ifs and whatever.

davidhalter:
Thank you man!

PR #18, opened 2017-10-27
tox.ini: prefer pytest without a dot

blueyed:


davidhalter:
Perfectionist ;-) Thanks!

PR #19, opened 2017-11-01
Support async for statement

robinro:
Generate correct definitions for 
`a = [func(x) async for x in foo()]`

Otherwise `x` will not be defined in the `func(x)` context.


davidhalter:
Good catch! Thanks!

PR #25, opened 2017-12-16
Make the documentation build reproducibility.

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that parso could not be built reproducibly.

Patch attached that uses SOURCE_DATE_EPOCH [1].

 [0] https://reproducible-builds.org/
 [1] https://reproducible-builds.org/specs/source-date-epoch/

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,7 +13,6 @@
 
 import sys
 import os
-import datetime
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
@@ -45,7 +44,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'parso'
-copyright = u'2012 - {today.year}, parso contributors'.format(today=datetime.date.today())
+copyright = u'parso contributors'
 
 import parso
 from parso.utils import version_info


lamby:
This has been filed in Debian as https://bugs.debian.org/884527

docs/conf.py
```diff
@@ -45,7 +46,8 @@
 
 # General information about the project.
 project = u'parso'
-copyright = u'2012 - {today.year}, parso contributors'.format(today=datetime.date.today())
+now = datetime.datetime.utcfromtimestamp(int(os.environ.get('SOURCE_DATE_EPOCH', time.time())))
+copyright = u'2012 - {now.year}, parso contributors'.format(now=now)
```
blueyed: Isn't the until-year in copyright information useless anyway?

davidhalter: How do you mean that?

davidhalter:
How did you even find this? I'm just wondering :)

blueyed:
Just something I'vevaguely remembered - maybe even from Ubuntu/Debian days.
Searching for "copyright year" brings up http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/ for example.

lamby:
Feel free to drop the year (which would obviously close this issue). I tend to not submit those particular requests as it's a "harder sell", prefering just to fix the existing data. :)

davidhalter:
I agree that having this discussion is probably harder than anything else :)

But in this case I guess @blueyed is right and we can probably remove the whole copyright thing. It doesn't make any sense. The copyright exists anyway and we're not really interested in protecting the documentation. If anything we'd be interested in protecting the code. Since the code is MIT licensed we obviously don't care about that either :)

So if you could just change the PR to remove everything, thank you :)

lamby:
@davidhalter Done.

davidhalter:
Thanks man!

PR #27, opened 2017-12-29
Added Recursive Child Printer

MohamedAlFahim:
This code allows the user to print all nested nodes within a parsed module. Good for debugging purposes too.


MohamedAlFahim:
I checked why some of the Travis tests failed, and saw that Travis itself had problems installing certain versions of Python.

davidhalter:
Sorry, this doesn't seem to be something that I want right now. I like the idea and I will think about integrating it into the parser tree in the future.

PR #29, opened 2017-12-30
Fix typos

lcolaholicl:
Just a trifling typo but better to be fixed.
* keyworda → keyword [parso/grammar.py]
* delete accidentally repeated `None.` [parso/tree.py]


davidhalter:
Thanks!

PR #30, opened 2017-12-30
tox: use older pytest only for py26/py33

blueyed:
Follow-up to https://github.com/davidhalter/parso/commit/73439d5863daea.


davidhalter:
Thanks!

PR #33, opened 2018-02-16
Fixes for tokenize

asmeurer:
Playing around with tokenize. Here are some fixes. 

How do you feel about making the second argument optional (defaulting to the current version)?

### DIFF
--- a/parso/python/tokenize.py
+++ b/parso/python/tokenize.py
@@ -222,7 +222,7 @@ class PythonToken(Token):
         return tok_name[self.type]
 
     def __repr__(self):
-        return ('TokenInfo(type=%s, string=%r, start=%r, prefix=%r)' %
+        return ('TokenInfo(type=%s, string=%r, start_pos=%r, prefix=%r)' %
                 self._replace(type=self._get_type_name()))
 
 


asmeurer:
Here's an annoying thing:

```
>>> list(parso.python.tokenize.tokenize("1 + 'a'", parse_version_string()))
[TokenInfo(type=NUMBER, string='1', start=(1, 0), prefix=''), TokenInfo(type=PLUS, string='+', start=(1, 2), prefix=' '), TokenInfo(type=STRING, string="'a'", start=(1, 4), prefix=' '), TokenInfo(type=ENDMARKER, string='', start=(1, 7), prefix='')]
>>> list(parso.python.tokenize.tokenize("1 + 'a", parse_version_string()))
[TokenInfo(type=NUMBER, string='1', start=(1, 0), prefix=''), TokenInfo(type=PLUS, string='+', start=(1, 2), prefix=' '), TokenInfo(type=ERRORTOKEN, string=" 'a", start=(1, 3), prefix=''), TokenInfo(type=ENDMARKER, string='', start=(1, 6), prefix='')]
```

The inclusion of the space in the ERRORTOKEN makes it more difficult to tell if the position being tokenized is part of a string or not. 

Also in general the lack of the `end` from the standard library makes it more annoying to deal with. I have to keep track of the previous token so I can check what token my `(row, col)` is part of.

asmeurer:
But I love that there are no (apparent) `TokenError` or `IndentationError`s!

davidhalter:
> How do you feel about making the second argument optional (defaulting to the current version)?

I think that makes sense.

> The inclusion of the space in the ERRORTOKEN makes it more difficult to tell if the position being tokenized is part of a string or not.

I would be fine with moving the prefix to the prefix variable. I think that makes more sense.

> Also in general the lack of the end from the standard library makes it more annoying to deal with. I have to keep track of the previous token so I can check what token my (row, col) is part of.

Why is `end_pos` not enough? It's an attribute on tokens.

asmeurer:
> Why is end_pos not enough? It's an attribute on tokens.

I didn't notice that. Why isn't it just part of the namedtuple?

davidhalter:
Because if it's part of the namedtuple it takes up space. It's also not always available and parso itself doesn't need it.

davidhalter:
Good.

And because you brought it up, I just fixed the issue you were running into with error leaf prefixes. I didn't change the default version stuff though, but that's probably less of an issue, just would comfortable sometimes.

PR #35, opened 2018-04-11
Added path param to load_grammar for loading custom grammar files

naueramant:


davidhalter:
This makes sense! Thank you!

Issue #34, opened 2018-03-08
Enable to disable LOG.Debug


giangm9:
I embed IPython in [blender](https://www.blender.org/), when I type a variable name and ' 'tab', parso's debug log shows up and this so annoying 
```
In [1]: armDEBUG:THREE:pickle loaded: /home/gl/tools/blender-2.79a-linux-glibc219-x86_64/2.79/scripts/modules/bpy_types.py
atureDEBUG:THREE:diff parser start
DEBUG:THREE:diff parser calculated
DEBUG:THREE:diff: line_lengths old: 1, new: 1
DEBUG:THREE:diff replace old[1:1] new[1:1]
DEBUG:THREE:parse_part from 1 to 1 (to 0 in part parser)
DEBUG:THREE:diff parser end
```
My current solution is remove log from source code file. I think there is must be a better way to disable Logs

davidhalter:
Disabling logs is possible with the builtin `logging` module. It's actually not that hard, just search for it. They might need to fix this inside the blender Python plugin, but that's all. 

You can also disable specific logs to avoid disabling everything (e.g. just disable parso).

This should usually work:

```
        root = logging.getLogger('parso')
        root.setLevel(logging.INFO)
```

Granitosaurus:
While setting level manually works I don't think it should be default on debug level. This is very unusual and creates all sorts of problems like the one described in OP.

This completely breaks `ptpython` shell as described in https://github.com/jonathanslenders/ptpython/issues/241

Granitosaurus:
:point_up: PR to fix this

PR #37, opened 2018-05-03
change loggers to use INFO as default logging level to prevent log spam

Granitosaurus:
See https://github.com/jonathanslenders/ptpython/issues/241 and https://github.com/davidhalter/parso/issues/34

### DIFF
--- a/parso/cache.py
+++ b/parso/cache.py
@@ -16,6 +16,7 @@ except:
 from parso._compatibility import FileNotFoundError
 
 LOG = logging.getLogger(__name__)
+LOG.setLevel('INFO')
 
 
 _PICKLE_VERSION = 30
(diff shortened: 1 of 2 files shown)


blueyed:
For reference, please see https://github.com/jonathanslenders/ptpython/issues/241#issuecomment-386207783 and the following discussion.
It should not be the job of a library to set its default logging level, but rather by / through the app.

amundhov:
Rather than tweak the logging levels, library code ought to add the NullHandler to the root logger.
That is

> import logging
> logging.getLogger('parso').addHandler(logging.NullHandler())


https://docs.python.org/3.10/howto/logging.html#configuring-logging-for-a-library

PR #46, opened 2018-08-02
Added mutmut to "used by" section of the docs

boxed:
Mutmut used to use baron but it's pretty much unmaintained. Moving to parso actually simplified my code quite a bit in addition to getting full python 3 support. So great and thank you a lot for this library!

### DIFF
--- a/docs/docs/usage.rst
+++ b/docs/docs/usage.rst
@@ -61,6 +61,8 @@ Used By
 -------
 
 - jedi_ (which is used by IPython and a lot of editor plugins).
+- mutmut_ (mutation tester)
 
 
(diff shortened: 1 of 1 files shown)


davidhalter:
Happy to see you are using it :)

~ Dave

boxed:
As in it tries to recover from invalid syntax and stuff? Well ok, I don’t need that for sure. What are the implications of having that on?

davidhalter:
Well, you will have no exceptions if the parser fails. Instead you will have ErrorNode and ErrorLeaf nodes in there. If you don't want error recovery, use `parso.parse(code, error_recovery=False)`.

boxed:
Thanks! Fixed.

PR #49, opened 2018-09-12
Add Python 3.7 stable to test matrix and update classifiers

michael-k:


.travis.yml
```diff
@@ -6,14 +6,14 @@ python:
   - 3.4
   - 3.5
   - 3.6
```
davidhalter: Why not just use 3.7 here? This should be working.

michael-k: No, it does not.  Python 3.7 requires OpenSSL 1.0.2+, which is not available on trusty.  See travis-ci/docs-travis-ci-com#1710 for the documentation adjustment and travis-ci/travis-ci#9069 as well as travis-ci/travis-ci#9815 for further discussion around the issue.

davidhalter: Thanks for the explanation and the PR!

PR #50, opened 2018-09-28
Use SVG badges

hugovk:
image

They look a bit nicer than the old PNGs:

image

And update some links to HTTPS.


davidhalter:
Great, thank you!

PR #38, opened 2018-05-10
Drop support for EOL Python 2.6 and 3.3

hugovk:
Python 2.6 and 3.3 are EOL and no longer receiving security updates (or any updates) from the core Python team.

They're also little used.

Here's the pip installs for parso from PyPI for last month:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 3.6            |  55.88% |        188,129 |
| 3.5            |  23.47% |         79,005 |
| 3.4            |  16.04% |         53,990 |
| 2.7            |   4.01% |         13,504 |
| 3.7            |   0.37% |          1,259 |
| 3.3            |   0.23% |            773 |
| 2.6            |   0.00% |             13 |
| 3.8            |   0.00% |              6 |

Source: `pypinfo --start-date -39 --end-date -10 --percent --pip --markdown parso pyversion`

PR #61, opened 2019-02-23
ci: Travis: use dist=xenial

blueyed:


blueyed:
py26 is not available there anymore: https://travis-ci.org/davidhalter/parso/jobs/497411068

Ref: https://github.com/davidhalter/parso/pull/38

davidhalter:
Just remove 2.6, next release will be without it.

davidhalter:
I'm also a bit confused about pypy2.7-6.0, can we not just use pypy2.7?

blueyed:
> I'm also a bit confused about pypy2.7-6.0, can we not just use pypy2.7?

No, that's the new name on Xenial.. Travis does not have an official list, but that works..

davidhalter:
Looks good, thanks!

PR #62, opened 2019-03-15
Fix order nondeterminism in DFA state generation.

DragonMinded:
Set traversal is unfortunately still non-deterministic in Python 3.6. This fixes non-deterministically choosing the wrong production in some cases when using a custom grammar with parso. I don't see the incorrect behavior with Python's default grammar however. This looks like it is probably a slight speed regression due to membership checking on list instead of set. I'm not sure if that's a problem in practice, however? Passes all tests that were passing pre-change in Py 2.7, 3.3 and 3.6.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -6,6 +6,7 @@ David Halter (@davidhalter) <davidhalter88@gmail.com>
 Code Contributors
 =================
 Alisdair Robertson (@robodair)
+Jennifer Taylor (@dragonminded)
 
 
 Code Contributors (to Jedi and therefore possibly to this library)
--- a/parso/pgen2/generator.py
+++ b/parso/pgen2/generator.py
@@ -71,7 +71,7 @@ class DFAState(object):
     different nonterminals.
     """
     def __init__(self, from_rule, nfa_set, final):
-        assert isinstance(nfa_set, set)
+        assert isinstance(nfa_set, (set, list))
         assert isinstance(next(iter(nfa_set)), NFAState)
         assert isinstance(final, NFAState)
         self.from_rule = from_rule
@@ -174,12 +174,12 @@ def _make_dfas(start, finish):
         assert isinstance(nfa_state, NFAState)
         if nfa_state in base_nfa_set:
             return
-        base_nfa_set.add(nfa_state)
+        base_nfa_set.insert(0, nfa_state)
         for nfa_arc in nfa_state.arcs:
             if nfa_arc.nonterminal_or_string is None:
                 addclosure(nfa_arc.next, base_nfa_set)
 
-    base_nfa_set = set()
+    base_nfa_set = []
     addclosure(start, base_nfa_set)
     states = [DFAState(start.from_rule, base_nfa_set, finish)]
     for state in states:  # NB states grows while we're iterating
@@ -188,7 +188,7 @@ def _make_dfas(start, finish):
         for nfa_state in state.nfa_set:
             for nfa_arc in nfa_state.arcs:
                 if nfa_arc.nonterminal_or_string is not None:
-                    nfa_set = arcs.setdefault(nfa_arc.nonterminal_or_string, set())
+                    nfa_set = arcs.setdefault(nfa_arc.nonterminal_or_string, [])
                     addclosure(nfa_arc.next, nfa_set)
 
         # Now create the dfa's with no None's in arcs anymore. All Nones have


davidhalter:
Just wondering about two things:

1. What's you're problem exactly? I would like to understand it, it would probably help if you had an example somewhere.
2. What are you using a custom grammar for?

Just interested :)

DragonMinded:
A teammate and I are parsing Python3 code (3.6 at the moment to be exact), but we are attempting to build a fully typed, higher-abstraction CST that resembles the python AST. We're experimenting with refactoring tools that present the end user with a tree hopefully matching Python's AST (https://greentreesnakes.readthedocs.io/en/latest/nodes.html) wherever possible so that complex rewrites don't need to be aware of syntax specifics (such as remembering to insert a comma if appending an argument to the end of an argument list). Its easier to fork the official grammar to create productions that can be transformed to this CST/AST hybrid than it is to parse the official grammar as-is and then walk/transform the resulting tree.

The production we were having trouble with specifically is `simple_stmt_partial: small_stmt (simple_stmt_middle_semicolon small_stmt)* [simple_stmt_trailing_semicolon] NEWLINE'` when parsing a test case containing "pass;continue". Both `simple_stmt_middle_semicolon` and `simple_stmt_trailing_semicolon` are defined as ";" exactly. Before this patch, on occasion the DFA would place `simple_stmt_middle_semicolon` after `simple_stmt_trailing_semicolon` and then the parser would parse "pass;" before complaining that "continue" was found but "NEWLINE" was expected. This appears to be due to using a set during construction of DFAs in "_make_dfas", coupled with the fact that both `simple_stmt_trailing_semicolon` and `simple_stmt_middle_semicolon` are the literal ";" so the parser sometimes traverses the LHS and thinks that it is a `simple_stmt_trailing_semicolon`, locking us into thinking the next token must be "NEWLINE".

davidhalter:
Can you add a test that fails without this change? I'm very willing to fix it, but I want to understand it a bit better.

DragonMinded:
Its not really possible to provide a test that fails, given that the order is indeterminate. At best, it would sometimes fail and pass other times (this is how I caught this on our own usecase). I ended up working around this problem internally by changing the grammar to not depend on ordered construction of DFAs.

bgw:
I spent a few hours staring at this, and I think that this is actually a shift/reduce conflict in disguise, but there's a few separate issues at play:

1. While some effort is made in `_calculate_first_plans` to detect ambiguities, it can only detect ambiguities in the first element of a production. The ambiguity in @DragonMinded's example occurs in the second element of `simple_stmt_partial`'s production.

2. pgen2 doesn't support any mechanism for explicitly resolving shift/reduce conflicts.

3. Because pgen2 uses sets to form the DFA, and the DFA's arcs are evaluated in their iteration order, and sets have a randomized iteration order in Python 3.3+ due to [hash randomization](https://docs.python.org/3/reference/datamodel.html#object.__hash__), the undetected shift-reduce conflict results in non-deterministic behavior, which is the worst possible outcome.

This pull request would fix issue 3, but I think issue 3 is just a symptom of issues 1 and 2.

To validate my assumptions about issues 1, 2, and 3, I crafted the following test program:

```python
from parso.pgen2.generator import generate_grammar, _dump_dfas

class Tokens:
    A = "A"
    B = "B"
    C = "C"

def dump(txt):
    grammar = generate_grammar(txt, Tokens)
    _dump_dfas(grammar.nonterminal_to_dfas["outer"])

# this will generate a DFA, though it will contain an ambiguity
dump("""
outer: A [inner] B C
inner: B C [inner]
""")

# this will fail to generate a DFA, because pgen2 detects the ambiguity
dump("""
outer: A rest
rest: [inner] B C
inner: B C [inner]
""")
```

The first `dump` call generates a DFA:

```
Dump of DFA for outer
  State 0 
    A -> 1
  State 1 
    inner -> 2
    B -> 3
  State 2 
    B -> 3
  State 3 
    C -> 4
  State 4 (final)
```

The DFA is reordered every time the program is run, but this should be acceptable behavior iff the DFA is unambiguous and equivalent. The DFA is equivalent but neither DFA is unambiguous.

```
Dump of DFA for outer
  State 0 
    A -> 1
  State 1 
    B -> 2
    inner -> 3
  State 2 
    C -> 4
  State 3 
    B -> 2
  State 4 (final)
```

It's not clear from either of these DFAs what transition from `State 1` should occur because `B` is the starting token for the `B` transition and the `inner` transition.

This shift/reduce conflict is validated by the second `dump` call, which uses an equivalent but rewritten grammar. It generates an exception because `_calculate_first_plans` is able to detect the ambiguity:

```
Traceback (most recent call last):
  File "test_dfa.py", line 21, in <module>
    """)
  File "test_dfa.py", line 9, in dump
    grammar = generate_grammar(txt, Tokens)
  File "/home/bgw/parso/parso/pgen2/generator.py", line 273, in generate_grammar
    _calculate_tree_traversal(rule_to_dfas)
  File "/home/bgw/parso/parso/pgen2/generator.py", line 309, in _calculate_tree_traversal
    _calculate_first_plans(nonterminal_to_dfas, first_plans, nonterminal)
  File "/home/bgw/parso/parso/pgen2/generator.py", line 353, in _calculate_first_plans
    % (nonterminal, t, nonterminal2, check[-1].from_rule)
ValueError: Rule rest is ambiguous; B is the start of the rule inner as well as rest.
```

My proposal is that we should extend pgen2 to detect these other forms of shift/reduce conflicts, and raise an exception.

More ideally, pgen2 should have a way to resolve shift/reduce conflicts, but would require more work, and isn't a hard requirement for parsing Python.

Thoughts?

davidhalter:
Fixed by #70.

Issue #63, opened 2019-03-21
tests fail under Python 3.8


carlwgeorge:
I noticed that Python 3.8 is at least on your mind due to #47 and #52.  However I discovered that the tests do not pass on that version.  Using the `python:3.8.0a2` container image from Docker Hub, I cloned this repo, ran `pip install -e .[testing]`, then `pytest`, and got many failures.

<details>
<summary>pytest output (click to expand)</summary>

```
root@tassadar:~/parso# pytest 
========================================================= test session starts ==========================================================
platform linux -- Python 3.8.0a2, pytest-4.3.1, py-1.8.0, pluggy-0.9.0
rootdir: /root/parso, inifile: pytest.ini
collected 937 items                                                                                                                    

parso/__init__.py .                                                                                                              [  0%]
parso/python/tree.py .                                                                                                           [  0%]
test/test_absolute_import.py ...                                                                                                 [  0%]
test/test_cache.py ..                                                                                                            [  0%]
test/test_diff_parser.py ............................................................                                            [  7%]
test/test_error_recovery.py ............                                                                                         [  8%]
test/test_file_python_errors.py .......                                                                                          [  9%]
test/test_fstring.py .................................                                                                           [ 12%]
test/test_get_code.py ......                                                                                                     [ 13%]
test/test_grammar.py .                                                                                                           [ 13%]
test/test_load_grammar.py ............                                                                                           [ 14%]
test/test_normalizer_issues_files.py ..................................                                                          [ 18%]
test/test_old_fast_parser.py ...............                                                                                     [ 19%]
test/test_param_splitting.py .........                                                                                           [ 20%]
test/test_parser.py ............................................................................................................ [ 32%]
.............                                                                                                                    [ 33%]
test/test_parser_tree.py ...................................................                                                     [ 39%]
test/test_pep8.py ...                                                                                                            [ 39%]
test/test_pgen2.py .......................x..................................................................................... [ 51%]
................................................................................................................................ [ 64%]
.................................................                                                                                [ 70%]
test/test_prefix.py .....................                                                                                        [ 72%]
test/test_python_errors.py ..F....F......FF..........FF........FFFFFFFFFFFFFFFFFFFFF.FFFFFFFFFFFFF................F............. [ 83%]
..................................FF.FFF.......F................................................                                 [ 93%]
test/test_tokenize.py ........................................                                                                   [ 97%]
test/test_utils.py ......................                                                                                        [100%]

=============================================================== FAILURES ===============================================================
[...]

PR #64, opened 2019-03-29
Test on Python 3.8

hroncok:


hroncok:
The tests fail, see https://github.com/davidhalter/parso/issues/63

davidhalter:
Thanks!

PR #69, opened 2019-05-06
Add type stubs. Fixes #68.

### LINKED ISSUE #68: Interest in shipping type annotations with the project?
Hi! Thanks for parso.

We are using parso in a project that uses Python PEP 484 type annotations and a typechecker, so we have written fairly comprehensive type stubs for Parso's API. We'd like to share this work with the community, if you are amenable. There are many options for how to do this:

1) Integrate type annotations directly into the codebase (need to use the comment-based syntax in order to preserve Python2 support). This is perhaps the best in terms of ensuring they stay up to date, but it depends if you want type hint comments in the code.

2) Add `.pyi` stub files next to the `.py` files in the repo. This is next-best for ensuring they stay up to date, but keeps them out of the code for those who don't care.

3) Keep them out of the parso repo entirely and include them in github.com/python/typeshed instead. This is the best option if you don't want to adopt any responsibility for maintaining them, and leave it up to the users of parso and typecheckers. But we won't do this either without your permission.

Let me know if any of those options sound acceptable to you, and which you prefer! Thanks again for your work on parso.

carljm:
This doesn't type the entire internal API, but it covers everything we use and I would guess that most users of Parso would need. LMK if you think anything is missing that really should be included, or if any of the types/signatures look wrong.

I did run mypy over these stubs and none of them raise any error on default settings.

### DIFF
--- a/parso/__init__.pyi
+++ b/parso/__init__.pyi
@@ -1,8 +1,8 @@
 from typing import Any, Optional, Union
 
-from parso.grammar import Grammar, load_grammar
-from parso.parser import ParserSyntaxError
-from parso.utils import python_bytes_to_unicode, split_lines
+from parso.grammar import Grammar as Grammar, load_grammar as load_grammar
+from parso.parser import ParserSyntaxError as ParserSyntaxError
+from parso.utils import python_bytes_to_unicode as python_bytes_to_unicode, split_lines as split_lines
 
 __version__: str = ...
 
--- a/parso/__init__.pyi
+++ b/parso/__init__.pyi
@@ -1,0 +1,19 @@
+from typing import Any, Optional, Union
+
+from parso.grammar import Grammar, load_grammar
+from parso.parser import ParserSyntaxError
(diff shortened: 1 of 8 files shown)


parso/__init__.pyi
```diff
@@ -0,0 +1,19 @@
+from typing import Any, Optional, Union
+
+from parso.grammar import Grammar, load_grammar
+from parso.parser import ParserSyntaxError
+from parso.utils import python_bytes_to_unicode, split_lines
```
davidhalter: For all of the imports in this paragraph: Don't you have to write `Grammar as Grammar, load_grammar as load_grammar`? Otherwise I don't think those are public. At least that's how I understand stubs and how I implemented them in Jedi.

carljm: Good call! It looks like our code is importing this stuff from its original location and not actually relying on this re-export, so we didn't catch this. Will fix.

davidhalter: Thanks a lot!

PR #70, opened 2019-05-07
Make pgen2's grammar ambiguity detection handle more cases (fixes #62)

bgw:
Under the old implementation,

```
outer: A [inner] B C
inner: B C [inner]
```

wouldn't get detected as the ambiguous grammar that it is, whereas

```
outer: A rest
rest: [inner] B C
inner: B C [inner]
```

would.

This would manifest itself as non-determinism in the DFA state generation. See the discussion #62 on for a full explanation.

This modifies the ambiguity detection to work on a broader class of issues, so it should now hopefully detect all cases where the given grammar is ambiguous.

At some point, we could extend this logic to allow developers to optionally set precedence of grammar productions, which could resolve ambiguities, but that's not a strict requirement for parsing python.

### DIFF
--- a/parso/pgen2/generator.py
+++ b/parso/pgen2/generator.py
@@ -309,13 +309,39 @@ def _calculate_tree_traversal(nonterminal_to_dfas):
             _calculate_first_plans(nonterminal_to_dfas, first_plans, nonterminal)
 
     # Now that we have calculated the first terminals, we are sure that
-    # there is no left recursion or ambiguities.
+    # there is no left recursion.
 
     for dfas in nonterminal_to_dfas.values():
         for dfa_state in dfas:
+            transitions = dfa_state.transitions
             for nonterminal, next_dfa in dfa_state.nonterminal_arcs.items():
                 for transition, pushes in first_plans[nonterminal].items():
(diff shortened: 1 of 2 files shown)


davidhalter:
Looks generally like a good idea. Please note that I currently don't have time for reviewing bigger pull requests. I also haven't answered to your other post for time management reasons (though I would really like that discussion!). This is a important for me and I will eventually come around to review it. Just give me some time.

bgw:
@davidhalter, that's fine. We're not blocked on this, and I wouldn't expect most people to run into this (because I assume most people won't be calling parso's pgen2 with a different grammar).

davidhalter:
BTW: If we wanted to make grammars like these work (ever, which we probably don't), something like this would be the way: https://en.wikipedia.org/wiki/GLR_parser. A GLL parser would probably do a good job with this, but I feel like it's not necessary at all for a tool like parso, you could just rewrite the grammar.

davidhalter:
BTW: e723b3e is changing the test structure a bit. I hope you don't mind. I feel like this is just quite a bit shorter and I'm starting to like parametrize more and more.

Issue #89, opened 2019-11-27
Incorrect parsing behaviour with assignment expressions


JarryShaw:
Recently, we encountered several issues with `parso` when implementing a backport compiler for assignment expressions.

1. Parsing f-strings with `:=` formatter produces incorrect SyntaxError
2. Parsing invalid use cases of assignment expressions do not raise SyntaxError

---

**NB**: all codes are running under Python 3.8.0; *actual behaviours* are running through `parso.parse(code, error_recovery=False, version='3.8')`.

## Case 1

**Source code**:

```python
f'{x:=5}'
```

**Expected behaviour**: valid, passes `=5` to formatter

**Actual behaviour**:

```pytb
Traceback (most recent call last):
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/parser.py", line 181, in _add_token
    plan = stack[-1].dfa.transitions[transition]
KeyError: ReservedString(:=)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/__init__.py", line 58, in parse
    return grammar.parse(code, **kwargs)
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/grammar.py", line 78, in parse
    return self._parse(code=code, **kwargs)
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/grammar.py", line 147, in _parse
    root_node = p.parse(tokens=tokens)
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/python/parser.py", line 82, in parse
    return super(Parser, self).parse(tokens)
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/parser.py", line 128, in parse
    self._add_token(token)
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/parser.py", line 187, in _add_token
    self.error_recovery(token)
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/python/parser.py", line 151, in error_recovery
    return super(Parser, self).error_recovery(token)
  File "/fakepath/.venv/lib/python3.8/site-packages/parso/parser.py", line 151, in error_recovery
    raise ParserSyntaxError('SyntaxError: invalid syntax', error_leaf)
parso.parser.ParserSyntaxError: ('SyntaxError: invalid syntax', <ErrorLeaf: TokenType(OP):':=', (1, 4)>)
```

## Case 2

[...]

PR #90, opened 2019-12-01
Fixing davidhalter/parso#89

JarryShaw:
> all changes are in `parso/python/errors.py`

* add utility function (`_get_namedexpr`) to extract all assignment expression (`namedexpr_test`) nodes
* add `is_namedexpr` parameter to `_CheckAssignmentRule._check_assignment` and special error message for assignment expression related assignment issues (*cannot use named assignment with xxx*)
* add assignment expression check to `_CompForRule` (*assignment expression cannot be used in a comprehension iterable expression*)
* add `_NamedExprRule` for special assignment expression checks
  - *cannot use named assignment with lambda*
  - *cannot use named assignment with subscript*
  - *cannot use named assignment with attribute*
  - and fallback general checks in `_CheckAssignmentRule._check_assignment`
* add `_ComprehensionRule` for special checks on assignment expression in a comprehension
  - *assignment expression within a comprehension cannot be used in a class body*
  - *assignment expression cannot rebind comprehension iteration variable 'xxx'*

Sample testing code:

```python
import parso

grammar = parso.load_grammar(version='3.8')

string = '''\
# Case 2
(lambda: x := 1)
# Case 3
(a[i] := x)
# Case 4
(a.b := c)
# Case 5
[i := 0 for i, j in range(5)]
[[(i := i) for j in range(5)] for i in range(5)]
[i for i, j in range(5) if True or (i := 1)]
[False and (i := 0) for i, j in range(5)]
# Case 6
[i+1 for i in (i := range(5))]
[i+1 for i in (j := range(5))]
[i+1 for i in (lambda: (j := range(5)))()]
# Case 7
class Example:
    [(j := i) for i in range(5)]
'''
module = grammar.parse(string, error_recovery=True)

errors = grammar.iter_errors(module)
print('\n'.join('[L%dC%d] %s' % (*error.start_pos, error.message) for error in errors))
```

Sample output:

```pytb
[L2C1] SyntaxError: cannot use named assignment with lambda
[L4C1] SyntaxError: cannot use named assignment with subscript
[L6C1] SyntaxError: cannot use named assignment with attribute
[L8C1] SyntaxError: assignment expression cannot rebind comprehension iteration variable 'i'
[L9C3] SyntaxError: assignment expression cannot rebind comprehension iteration variable 'i'
[L10C36] SyntaxError: assignment expression cannot rebind comprehension iteration variable 'i'
[L11C12] SyntaxError: assignment expression cannot rebind comprehension iteration variable 'i'
[L13C15] SyntaxError: assignment expression cannot be used in a comprehension iterable expression
[L14C15] SyntaxError: assignment expression cannot be used in a comprehension iterable expression
[L15C24] SyntaxError: assignment expression cannot be used in a comprehension iterable expression
[L18C6] SyntaxError: assignment expression within a comprehension cannot be used in a class body
```

### DIFF
--- a/parso/python/errors.py
+++ b/parso/python/errors.py
@@ -125,6 +125,20 @@ def _get_for_stmt_definition_exprs(for_stmt):
     return list(_iter_definition_exprs_from_lists(exprlist))
 
 
+def _get_namedexpr(node):
+    """Get assignment expression if node contains."""
+    namedexpr_list = list()
+
+    def fetch_namedexpr(node):
+        if node.type == 'namedexpr_test':
+            namedexpr_list.append(node)
+        if hasattr(node, 'children'):
+            [fetch_namedexpr(child) for child in node.children]
+
+    fetch_namedexpr(node)
+    return namedexpr_list
+
+
 class _Context(object):
     def __init__(self, node, add_syntax_error, parent_context=None):
         self.node = node
@@ -883,7 +897,7 @@ class _FStringRule(SyntaxRule):
 
 
 class _CheckAssignmentRule(SyntaxRule):
-    def _check_assignment(self, node, is_deletion=False):
+    def _check_assignment(self, node, is_deletion=False, is_namedexpr=False):
         error = None
         type_ = node.type
         if type_ == 'lambdef':
@@ -907,9 +921,9 @@ class _CheckAssignmentRule(SyntaxRule):
                         # This is not a comprehension, they were handled
                         # further above.
                         for child in second.children[::2]:
-                            self._check_assignment(child, is_deletion)
+                            self._check_assignment(child, is_deletion, is_namedexpr)
                     else:  # Everything handled, must be useless brackets.
-                        self._check_assignment(second, is_deletion)
+                        self._check_assignment(second, is_deletion, is_namedexpr)
         elif type_ == 'keyword':
             if self._normalizer.version < (3, 8):
                 error = 'keyword'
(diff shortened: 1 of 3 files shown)


JarryShaw:
Wait... just realise there're some bugs when dealing with comprehensions... 🤦‍♂

parso/python/errors.py
```diff
@@ -1008,3 +1033,82 @@ def is_issue(self, for_stmt):
...
+                    elif first_child.value == '.':
+                        # (a.b := c)
+                        self.add_issue(namedexpr_test, message='cannot use named assignment with attribute')
+        else:
+            self._check_assignment(first, is_namedexpr=True)
+
+
+@ErrorFinder.register_rule(type='testlist_comp')
+@ErrorFinder.register_rule(type='dictorsetmaker')
+class _ComprehensionRule(SyntaxRule):
```
davidhalter: Can we replace this with parent accesses in `_NamedExprRule`? There's also `search_ancestor` that might help with it. Not 100% sure if it works, but I feel like it's nicer than to scan for named exprs below.

JarryShaw: Okay, it's all done.

JarryShaw:
I have revised the implementation and added some failing & valid examples.

davidhalter:
I refactored a few small things and then I identified some small issues.

I first wanted to let you do the changes, but quickly realized that it wasn't easy to even tell you what was wrong. At first I was just trying to change your implementation, then realized you did most things right and then realized that you didn't catch a few edge cases that I now implemented. There's probably still a few cases we haven't caught, but I think this is now at least much better than before (and some cases might be so exotic and stupid that nobody will ever actually catch them).

PR #92, opened 2019-12-04
Fix trailing comma error

### LINKED ISSUE #91: False Positive: SyntaxError: trailing comma not allowed without surrounding parentheses
Hi,
I encountered a False positive about _trailing comma not allowed without surrounding parentheses_. I've made imports with trailing comma *with surrounding parenthesis* but the parser raises this error, even if it's valid.

```python
import parso

string = """
from hashlib import (
    md5,
    sha256,
)
"""

grammar = parso.load_grammar(version='3.8')
module = grammar.parse(string, error_recovery=True)

print(grammar.iter_errors(module)[0].message)
#SyntaxError: trailing comma not allowed without surrounding parentheses
```

RobinFrcd:
Fixes #91

### DIFF
--- a/test/test_python_errors.py
+++ b/test/test_python_errors.py
@@ -307,3 +307,15 @@ def test_invalid_fstrings(code, message):
     """
     error, = _get_error_list(code, version='3.6')
     assert message in error.message
+
+
+@pytest.mark.parametrize(
+    'code', [
+        "from foo import (\nbar,\n rab,\n)",
+        "from foo import (bar, rab, )",
+    ]
+)
+def test_trailing_comma(code):
+    errors = _get_error_list(code)
+    assert not errors
+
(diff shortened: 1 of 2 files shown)


test/test_python_errors.py
```diff
@@ -41,6 +41,18 @@ def test_python_exception_matches(code):
...
+def test_trailing_comma():
+    codes = [
+        "from foo import (\nbar,\n rab,\n)",
+        "from foo import (bar, rab, )",
+    ]
+
+    wanted = list()
+    for code in codes:
+        errors = _get_error_list(code)
+        assert errors == wanted
```
davidhalter: Please rewrite this to:

```
@pytest.mark.parametrize(
    'code', [
        "from foo import (\nbar,\n rab,\n)",
        "from foo import (bar, rab, )",
    ]
)
def test_trailing_comma(code):
    errors = _get_error_list(code)
    assert not errors
```

davidhalter: I will merge after that.

davidhalter:
Thanks for your contribution!

davidhalter:
Thanks a lot!

PR #97, opened 2020-01-02
Revision on assignment errors

JarryShaw:
* added rule for `__debug__` (should be a keyword)
 * reviewed error messages (26/33 not fully tested)
 * added new failing samples


JarryShaw:
Btw, I also noticed that in **Python 2.6**:

```python
>>> None = 1
  File "<stdin>", line 1
SyntaxError: assignment to None
```

I didn't include this special case in the PR.

davidhalter:
We are going to drop Python 2.6 soon anyway.

parso/python/errors.py
```diff
@@ -961,12 +975,26 @@ def _check_assignment(self, node, is_deletion=False, is_namedexpr=False):
...
               or type_ in ('term', 'factor')):
             error = 'operator'
+        elif type_ == 'name':
+            if node.value == '__debug__':  # should be a keyword for all versions
+                if (2, 7) < self._normalizer.version < (3, 8):
+                    error = 'keyword'
+                else:
+                    error = str(node.value)
+                not_debug = False
+                is_keyword = True
```
davidhalter: IMO this belongs to `_NameChecks`, because there we already check for `__debug__`.

davidhalter: It might be worth patching is_definition.

parso/python/errors.py
```diff
@@ -961,12 +975,26 @@ def _check_assignment(self, node, is_deletion=False, is_namedexpr=False):
...
-            if is_namedexpr:
+            if is_namedexpr and not_debug:
+                # c.f. CPython bpo-39176, should be changed in next release
+                # message = 'cannot use assignment expressions with %s' % error
                 message = 'cannot use named assignment with %s' % error
             else:
-                cannot = "can't" if self._normalizer.version < (3, 8) else "cannot"
+                _version = self._normalizer.version
+                if _version[0] < 3:  # Python 2.*
+                    cannot = ("can not" if _version < (2, 7) else "cannot") if is_keyword else "can't"
```
davidhalter: IMO ignore Python2.6 here. I will just drop support for it. Nobody uses that anymore.

davidhalter: Also I wouldn't add the keyword variable. The messages don't need to match exactly as long as they are at least there. If it tests cause any trouble, add an additional case to  `_get_actual_exception` in `test_python_errors.py`.

JarryShaw: Okay, I removed all these crazy cases and just patched the "actual" exceptions.

parso/python/errors.py
JarryShaw: Just patched, I suppose it's working fine :)

parso/python/errors.py
JarryShaw: Ah... seems you just dropped Python 2... lol

davidhalter: > Ah... seems you just dropped Python 2... lol

All because of you, hahaha :D

davidhalter:
Thanks a lot for looking into this! :) I have a hard time following all the changes, especially because I don't really use the output of the errors anywhere.

PR #100, opened 2020-01-09
Revision on fstring issues

JarryShaw:
* f-string expression part cannot include a backslash
 * failing example `f"{'\n'}"` for tests

Yes, me again :)


davidhalter:
Thanks!

PR #101, opened 2020-02-14
Add venv to .gitignore

Jma353:
Similar to my [commit to `jedi` of a similar form.](https://github.com/davidhalter/jedi/commit/d42d3f45f08e06f3893e6e948c332cc0e52a2ca9)

For Github contributions to different libraries, I like keeping different Python `virtualenv`'s.  `venv` is the most common name for such a directory, so proactively `.gitignore` it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ parso.egg-info/
 /.cache/
 /.pytest_cache
 test/fuzz-redo.pickle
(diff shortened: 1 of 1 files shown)


blueyed:
Not `.venv`?
Anyway, I think you should add it to your global gitignore (also) :)

PR #108, opened 2020-05-17
approve hit msg when python version is unsupported

WinChua:
currently, when the python version used is not supported, it will raise "Python version None is currently not supported.". 

Maybe it could be changed to raise NotImplement("Python version x.x is currently not supported")


davidhalter:
Makes sense, thanks!

PR #109, opened 2020-05-21
Correctly parse 2-digit minor versions (py3.10)

isidentical:
With the release of py3.10, parso started to fail parsing it's version number
```
Traceback (most recent call last):
  File "/home/isidentical/syntax_test_suite/unimportz.py", line 11, in <module>
    from unimport.session import Session
  File "/home/isidentical/.venvs/master/lib/python3.10/site-packages/unimport/session.py", line 7, in <module>
    from unimport.refactor import refactor_string
  File "/home/isidentical/.venvs/master/lib/python3.10/site-packages/unimport/refactor.py", line 1, in <module>
    import libcst as cst
  File "/home/isidentical/.venvs/master/lib/python3.10/site-packages/libcst/__init__.py", line 187, in <module>
    from libcst._parser.entrypoints import parse_expression, parse_module, parse_statement
  File "/home/isidentical/.venvs/master/lib/python3.10/site-packages/libcst/_parser/entrypoints.py", line 25, in <module>
    _DEFAULT_PARTIAL_PARSER_CONFIG: PartialParserConfig = PartialParserConfig()
  File "<string>", line 8, in __init__
  File "/home/isidentical/.venvs/master/lib/python3.10/site-packages/libcst/_parser/types/config.py", line 121, in __post_init__
    parsed_python_version = parse_version_string(
  File "/home/isidentical/.venvs/master/lib/python3.10/site-packages/libcst/_parser/parso/utils.py", line 218, in parse_version_string
    return _parse_version(version)
  File "/home/isidentical/.venvs/master/lib/python3.10/site-packages/libcst/_parser/parso/utils.py", line 187, in _parse_version
    raise ValueError(
ValueError: The given version is not in the right format. Use something like "3.2" or "3".
```

PR #110, opened 2020-05-21
Quick fix about invalid version test

isidentical:
I missed this when I did the #109 (I wasn't able to run tests since there wasn't a 3.10 grammar).


davidhalter:
Question is more like: Why did I mess it when the tests were clearly failing...

PR #111, opened 2020-05-21
Adapt Python3.9 errors on multiple star target

isidentical:
In Python3.9, the message "two starred expression in ..." changed
to "multiple starred expression in ...", with python/cpython#19168


davidhalter:
Thanks!

PR #112, opened 2020-05-21
Allow 'any' expression on decorators, PEP 614

isidentical:
PEP 614 (implemented with 3.9) now allows any expression on decorators https://www.python.org/dev/peps/pep-0614/


davidhalter:
Thanks!

PR #115, opened 2020-05-22
Support finally in continue on 3.8+

isidentical:
Thanks to [bpo-32489](https://bugs.python.org/issue32489) and sadly
for rejection of my [PEP 601](https://www.python.org/dev/peps/pep-0601/)
finally in continue is supported in 3.8+. I checked the blame and looks
like there was already a commit for the same subject, but that only
changes the test and not actually changes the checker (dfe7fba08e993b4c0f2e0af75f81e787fc399297)

### DIFF
--- a/parso/python/errors.py
+++ b/parso/python/errors.py
@@ -449,7 +449,11 @@ class _ContinueChecks(SyntaxRule):
                 in_loop = True
             if block.type == 'try_stmt':
                 last_block = block.children[-3]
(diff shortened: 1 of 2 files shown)


davidhalter:
Great, thanks!

davidhalter:
HOW DO I NOT SEE THAT THE CI DOESN'T PASS...

Seriously. This is just stupid on my part.  Can you look into it @isidentical?

PR #116, opened 2020-05-22
Raise violation on starred expressions where the child is a boolean/none

isidentical:


isidentical:
@davidhalter apparently I broke the CI, for the second time. Terribly sorry for this, this PR also includes a fix for that breakage.

davidhalter:
> @davidhalter apparently I broke the CI, for the second time. Terribly sorry for this, this PR also includes a fix for that breakage.

No worries, it just has to work in the end :)

------

I think this should be done here: https://github.com/davidhalter/parso/blob/aa146f01246050c002403e65cb6d99fc40bc540d/parso/python/errors.py#L921

The error message is already generated, but I think the star syntax is probably not handled properly.

isidentical:
Oh, I didn't see that. Thanks for the tip, I think it is much more simpler now @davidhalter

PR #119, opened 2020-05-23
Check all arguments for unparenthesized generator expressions

isidentical:
Previously only the first argument on the argument list checked
against the generator expressions, now all argumnets are controlled.

### DIFF
--- a/parso/python/errors.py
+++ b/parso/python/errors.py
@@ -779,62 +779,59 @@ class _ArglistRule(SyntaxRule):
             return "Generator expression must be parenthesized"
 
     def is_issue(self, node):
-        first_arg = node.children[0]
-        if first_arg.type == 'argument' \
-                and first_arg.children[1].type in _COMP_FOR_TYPES:
-            # e.g. foo(x for x in [], b)
-            return len(node.children) >= 2
-        else:
-            arg_set = set()
-            kw_only = False
-            kw_unpacking_only = False
-            is_old_starred = False
-            # In python 3 this would be a bit easier (stars are part of
-            # argument), but we have to understand both.
-            for argument in node.children:
-                if argument == ',':
-                    continue
+        arg_set = set()
+        kw_only = False
+        kw_unpacking_only = False
+        is_old_starred = False
+        # In python 3 this would be a bit easier (stars are part of
+        # argument), but we have to understand both.
+        for argument in node.children:
+            if argument == ',':
+                continue
 
-                if argument in ('*', '**'):
-                    # Python < 3.5 has the order engraved in the grammar
-                    # file.  No need to do anything here.
-                    is_old_starred = True
-                    continue
-                if is_old_starred:
-                    is_old_starred = False
-                    continue
+            if argument in ('*', '**'):
+                # Python < 3.5 has the order engraved in the grammar
+                # file.  No need to do anything here.
+                is_old_starred = True
+                continue
+            if is_old_starred:
+                is_old_starred = False
+                continue
 
-                if argument.type == 'argument':
-                    first = argument.children[0]
-                    if first in ('*', '**'):
-                        if first == '*':
-                            if kw_unpacking_only:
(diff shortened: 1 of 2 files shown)


isidentical:
The diff might look a bit messy, but I had to dedent all the code under the else to prevent 2 seperate iteration of arguments. If you'd like I can preserve the body of the else and do 2-iteration.

test/test_python_errors.py
```diff
@@ -375,3 +375,15 @@ def test_repeated_kwarg():
         _get_error_list("f(q=1, q=2)", version="3.9")[0].message
         == "SyntaxError: keyword argument repeated: q"
     )
+
+
+@pytest.mark.parametrize(
+    'source', [
+        'a(a for a in b,)'
```
davidhalter, line 382: I think this case should actually pass:

```
>>> str(a for a in [1],)
'<generator object <genexpr> at 0x7fecc48320a0>'
```

You should probably add a second param `has_errors` and parametrize with that as well. Please also add a simple case like `a(a for a in b)`.

Parso is sometimes not tested extremely well. The reason for this is that for a very long time Parso was part of Jedi and didn't really have its own tests. Nowadays there are at least a few tests that help, but they are not extensive at all.

isidentical: interesting.
```
 $ python
Python 3.8.0+ (heads/3.8:b9e5547f58, Nov 28 2019, 19:18:03) 
[GCC 9.2.1 20191008] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a(a for a in b,)
  File "<stdin>", line 1
SyntaxError: Generator expression must be parenthesized
>>> 
```

isidentical:
```
 $ python3.10
Python 3.10.0a0 (heads/bpo-xxxxx:f2947e354c, May 21 2020, 18:54:57) 
[GCC 9.2.1 20191008] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> str(a for a in [1],)
  File "<stdin>", line 1
    str(a for a in [1],)
        ^
SyntaxError: Generator expression must be parenthesized
```

isidentical: Looks like a python2 issue, will add a condition

davidhalter: BTW: Just in general, I don't think 3.5/2.7 support is important anymore. So don't put in too much effort. This is just something I found by accident, because 3.5 was open. IMO I wouldn't even fix it, because people that do that will soon be screwed anyway :)

I guess I was more worried that `str(a for a in [1])` would also not pass.

isidentical: > I guess I was more worried that str(a for a in [1]) would also not pass.
```
>>> get_errors("str(a for a in b)")
[]
```

parso/python/errors.py
```diff
@@ -779,62 +779,68 @@ def message(self):
...
+                first = argument.children[0]
+                if (
+                    argument.children[1].type in _COMP_FOR_TYPES
+                    and len(node.children) >= 2
+                ):
+                    if self._normalizer.version >= (3, 7):
+                        return True
+                    elif len(node.children) == 2 and node.children[1] == ",":
+                        # trailing comma allowed until 3.7
+                        pass
```
davidhalter, line 811: Sorry for the noise, but I feel like this extra case just makes it more complicated without an extra benefit.

Would you mind just ignoring it again? :) Sorry, but I really didn't know that this was only < 3.7. I thought it was a general issue.

davidhalter: If you think that this is more correct, I'm fine with that as well, but IMO it really helps nobody to have this distinction for Python 3.5/3.6.

isidentical: I agree, will revert

davidhalter:
Alright. I think my confusion really stemmed from why this would be working, that's why I even started testing. Until I finally found the `return True`, at that point everything made sense :). Thanks!

PR #121, opened 2020-05-24
Clear inactive cache files

### LINKED ISSUE #120: Parso Cache Garbage Collection
Currently people using parso might be using a lot of disk space after a while. `~/.cache/parso` might get quite big. (or whatever it is on other OS's, see `parso/cache.py`.

IMO we should probably clean up in `parso.cache.save_module`. That function should just check "once a day" if files are older than a month. In that case they should be deleted. It should be fine to delete stuff that we might parse again, since the cache is not extremely important and just there to speed up some things.

As a small additional thing, we should probably use `scandir`, see also https://github.com/davidhalter/jedi/blob/7fd5c8af8ff04711599fea10ca8babe51b280464/jedi/_compatibility.py#L29-L47 for Python 2.7/3.5. However we could also just remove Python 2.7/3.5 support for this feature. I don't plan on maintaining it for a long time.

isidentical:
Cache files that weren't accessed in the last 30 days will be automatically
garbage collected. This collection happens when the `save_module` is called
via a lock system that would make it happen only one time per day. Resolves #120

### DIFF
--- a/parso/cache.py
+++ b/parso/cache.py
@@ -100,8 +100,7 @@ def _get_cache_clear_lock(cache_path = None):
     collection once a day (can be configured in _CACHE_CLEAR_THRESHOLD).
     """
     cache_path = cache_path or _get_default_cache_path()
-
-    return FileIO(os.path.join(cache_path, "{}-cache-lock".format(_VERSION_TAG)))
+    return FileIO(os.path.join(cache_path, "PARSO-CACHE-LOCK"))
 
 
 parser_cache = {}
@@ -214,7 +213,8 @@ def clear_inactive_cache(
     cache_path=None,
     inactivity_threshold=_CACHED_FILE_MAXIMUM_SURVIVAL,
 ):
-    cache_path = _get_default_cache_path()
+    if cache_path is None:
+        cache_path = _get_default_cache_path()
     if not os.path.exists(cache_path):
         return False
     for version_path in os.listdir(cache_path):
@@ -226,7 +226,10 @@ def clear_inactive_cache(
                 file.stat().st_atime + _CACHED_FILE_MAXIMUM_SURVIVAL
                 <= time.time()
             ):
-                os.remove(file.path)
+                try:
+                    os.remove(file.path)
+                except OSError: # silently ignore all failures
+                    continue
     else:
         return True
 
@@ -238,8 +241,13 @@ def _remove_cache_and_update_lock(cache_path = None):
         clear_lock_time is None # first time
         or clear_lock_time + _CACHE_CLEAR_THRESHOLD <= time.time()
     ):
-        if clear_inactive_cache(cache_path = cache_path):
-            lock._touch()
+        if not lock._touch():
+            # First make sure that as few as possible other cleanup jobs also
+            # get started. There is still a race condition but it's probably
+            # not a big problem.
+            return False
+
+        clear_inactive_cache(cache_path = cache_path)
 
 def _get_hashed_path(hashed_grammar, path, cache_path=None):
     directory = _get_cache_directory_path(cache_path=cache_path)
--- a/parso/cache.py
+++ b/parso/cache.py
@@ -13,7 +13,8 @@ try:
 except:
     import pickle
 
-from parso._compatibility import FileNotFoundError
+from parso._compatibility import FileNotFoundError, scandir
+from parso.file_io import FileIO
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


davidhalter:
I might need a few days to merge this (want to review it well and I'm unfortunately not a student anymore :)).

isidentical:
> I might need a few days to merge this (want to review it well and I'm unfortunately not a student anymore :)).

No problem at all, in the meanwhile, I need to make this patch working on 3.5< with my force-pushes to use this CI.

isidentical:
All green

parso/cache.py
```diff
@@ -187,6 +210,37 @@ def clear_cache(cache_path=None):
...
+
+def _remove_cache_and_update_lock(cache_path = None):
+    lock = _get_cache_clear_lock(cache_path=cache_path)
+    clear_lock_time = lock.get_last_modified()
+    if (
+        clear_lock_time is None # first time
+        or clear_lock_time + _CACHE_CLEAR_THRESHOLD <= time.time()
+    ):
+        if clear_inactive_cache(cache_path = cache_path):
+            lock._touch()
```
davidhalter, line 242: IMO there shouldn't be an if. You should probably do it like this:

```
    def _touch(self):
        try:
            os.utime(self.path, None)
        except FileNotFoundError:
            try:
                file = open(self.path, 'a')
                file.close()
            except OSError:  # TODO Maybe log this?
                return False
        return True
    
# First make sure that as few as possible other cleanup jobs also get started. There is still a race condition but it's probably not a big problem.
if not lock._touch():
    return

clear_inactive_cache(cache_path = cache_path)
```

parso/cache.py
```diff
@@ -187,6 +210,37 @@ def clear_cache(cache_path=None):
     parser_cache.clear()
 
 
+def clear_inactive_cache(
+    cache_path=None,
+    inactivity_threshold=_CACHED_FILE_MAXIMUM_SURVIVAL,
+):
+    cache_path = _get_default_cache_path()
```
davidhalter, line 217: I think we should iterate through all the folders in `~/.cache/parso` instead of just the specific version. This also means that we might need to change something about `_get_cache_clear_lock` and that it should be "global".

davidhalter: The reasoning for this is mostly that people will upgrade Python versions and in that case the old caches will just stick around forever, which makes no sense. For example in my case I have the following folders:

```
CPython-27-30  CPython-33-30  CPython-34-33  CPython-35-33  CPython-36-33  CPython-37-33
CPython-27-31  CPython-34-30  CPython-35-30  CPython-36-30  CPython-37-30  CPython-38-31
CPython-27-32  CPython-34-31  CPython-35-31  CPython-36-31  CPython-37-31  CPython-38-32
CPython-27-33  CPython-34-32  CPython-35-32  CPython-36-32  CPython-37-32  CPython-38-33
```

The whole thing uses 3.5 GB. :)

parso/cache.py
```diff
@@ -187,6 +210,37 @@ def clear_cache(cache_path=None):
...
+    for version_path in os.listdir(cache_path):
+        version_path = os.path.join(cache_path, version_path)
+        if not os.path.isdir(version_path):
+            continue
+        for file in scandir(version_path):
+            if (
+                file.stat().st_atime + _CACHED_FILE_MAXIMUM_SURVIVAL
+                <= time.time()
+            ):
+                os.remove(file.path)
```
davidhalter, line 229: I think here we should catch OSError or like at least the errors and just ignore it. 

We can also be more strict and just ignore the errors that are caused if the user doesn't own the file. I guess that would be PermissionError?!

davidhalter:
Great work altogether. I just think we should be a bit more lenient with errors.

isidentical:
Thanks for the review @davidhalter, I'll try to look these next week

parso/cache.py
```diff
@@ -187,6 +209,45 @@ def clear_cache(cache_path=None):
     parser_cache.clear()
 
 
+def clear_inactive_cache(
+    cache_path=None,
+    inactivity_threshold=_CACHED_FILE_MAXIMUM_SURVIVAL,
+):
+    cache_path = _get_default_cache_path()
```
davidhalter, line 216: There should probably be an `if cache_path is None` before this.

davidhalter:
And as a minor nitpick: I actually prefer if people don't force push in pull requests. It's very hard to understand what has been changed. I can still squash if I really need to when I merge.

isidentical:
I try to keep the commit log clean, like for the last force pushes I was using trying to fix a CI failure on the same commit so I didn't want to create new commits just for that. But I respect your preference, so try to avoid them.

davidhalter:
As I said, it's a nitpick, feel free to do it however you want to. I feel it's obviously easier to spot the differences in PRs, but I can still work around that.

I also personally think that commit logs don't matter nearly as much as people think they do. If I want to write a longer text in a commit log I usually abort and make the exact comment in code. But here opinions tend to vary a lot and I'm probably not even part of the majority. In the end all I want is readable concise error-free code.

davidhalter:
(and as a small addition, I like to actually have the whole history of changes. For example I think your first iteration was really interesting, because it was simply another approach, but as I said, a lot of people don't feel that way)

Feel free to merge when you think you're done.

PR #125, opened 2020-05-25
Add Python3.10 grammar

isidentical:
What do you think about this feature @davidhalter, I'm not super sure but I believe there should be at least a warning to indicate these versions might not be fully parseable by parso. (Also this patch introduces a 3.10 grammar, which is basically a copy of 3.9)

### DIFF
--- a/parso/python/grammar310.txt
+++ b/parso/python/grammar310.txt
@@ -1,0 +1,171 @@
+# Grammar for Python
+
+# NOTE WELL: You should also follow all the steps listed at
+# https://devguide.python.org/grammar/
+
+# Start symbols for the grammar:
+#       single_input is a single interactive statement;
+#       file_input is a module or sequence of commands read from an input file;
+#       eval_input is the input for the eval() functions.
+# NB: compound_stmt in single_input is followed by extra NEWLINE!
+single_input: NEWLINE | simple_stmt | compound_stmt NEWLINE
+file_input: stmt* ENDMARKER
+eval_input: testlist NEWLINE* ENDMARKER
+
+decorator: '@' namedexpr_test NEWLINE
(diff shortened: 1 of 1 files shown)


davidhalter:
I'm actually opposed to raising a warning. This will probably just be annoying for some people working on the dev versions of Python (i.e. CPython devs). It's pretty obvious that we don't support a grammar if the language is not fully specified.

Adding the 3.10 grammar is of course fine, but I'd rather not merge the warning part. Python doesn't change as much these days, so even the 3.8 grammar for 3.9 should be mostly fine. If you want to solve this issue, we should probably document it well in the README/rtd.

isidentical:
The initial reason that I added these is we should tell somehow that we are not ready to parse non LL(1) (actually this is not exactly true, we are already parsing some parts that are not LL(1) but you know what I mean for things like multi-line with statements etc.)

davidhalter:
I don't understand what you mean. Can you elaborate?

isidentical:
> I don't understand what you mean. Can you elaborate?

https://bugs.python.org/issue12782

isidentical:
Actually, after some thinking I agree with you that this will only annoy people for no gain. I'll update this PR for only 3.10 grammar addition

davidhalter:
We are basically only parsing things that are LL(1). I guess I understand a bit better now.

I think my argument still holds. Almost all things in 3.10 will be the same as in 3.9. For the few things that are/will be(?) different we have to find solutions. The warning will just be annoying for people working with parso.  I think the key here is documentation.

davidhalter:
Ok, I guess my comment was a bit too late :)

isidentical:
> Ok, I guess my comment was a bit too late :)

Yup :)

PR #135, opened 2020-06-04
Improve handling of starred expression on different contexts

### LINKED ISSUE #133: Syntax error when using starred expressions in list and set displays
I'm not very familiar with how jedi works, I've just been using a vim plugin that uses jedi for python auto-completions.

But I think I've noticed a bug with jedi flagging starred expressions in list and set displays as syntax errors,
```python
>>> jedi.Script("[*(1,2,3)]").get_syntax_errors()
[<SyntaxError from=(1, 1) to=(1, 9)>]
>>> jedi.Script("{*(1,2,3)}").get_syntax_errors()
[<SyntaxError from=(1, 1) to=(1, 9)>]
```

If the starred expression is not the only item in the list, or if there is a trailing comma, there are no syntax errors,
```python
>>> jedi.Script("[*(1,2,3),]").get_syntax_errors()
[]
>>> jedi.Script("[*(1,2,3), *(4,5,6)]").get_syntax_errors()
[]
>>> jedi.Script("[0, *(1,2,3)]").get_syntax_errors()
[]
```

However, it seems like using a starred expression in a set always gives a syntax error,
```python
>>> jedi.Script("{*(1,2,3),}").get_syntax_errors()
[<SyntaxError from=(1, 1) to=(1, 9)>]
>>> jedi.Script("{*(1,2,3), *(4,5,6)}").get_syntax_errors()
[<SyntaxError from=(1, 1) to=(1, 9)>]
>>> jedi.Script("{0, *(4,5,6)}").get_syntax_errors()
[<SyntaxError from=(1, 4) to=(1, 12)>]
```

Im using Python 3.8.2,
```python
>>> jedi.Script("")
<Script: None <SameEnvironment: 3.8.2 in /usr>>
```
and my jedi version is
```python
>>> jedi.__version__
'0.17.0'
```

Unpacking in lists and sets should be allowed since Python 3.5 I think, see [PEP 448](https://www.python.org/dev/peps/pep-0448/).

isidentical:
Resolves #133

### DIFF
--- a/test/test_python_errors.py
+++ b/test/test_python_errors.py
@@ -413,11 +413,4 @@ def test_unparenthesized_genexp(source, no_errors):
     ]
 )
 def test_starred_expr(source, no_errors):
-    assert (
-        any(
-            "starred assignment target must be in a list or tuple"
-            in error.message
-            for error in _get_error_list(source)
-        )
-        ^ no_errors
-    )
+    assert bool(_get_error_list(source, version="3")) ^ no_errors
--- a/test/test_python_errors.py
+++ b/test/test_python_errors.py
@@ -391,3 +391,26 @@ def test_repeated_kwarg():
 )
 def test_unparenthesized_genexp(source, no_errors):
     assert bool(_get_error_list(source)) ^ no_errors
+
+@pytest.mark.parametrize(
+    ('source', 'no_errors'), [
+        ('*x = 2', False),
+        ('(*y) = 1', False),
+        ('((*z)) = 1', False),
+        ('a, *b = 1', True),
+        ('a, *b, c = 1', True),
+        ('a, (*b), c = 1', True),
+        ('a, ((*b)), c = 1', True),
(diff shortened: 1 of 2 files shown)


test/test_python_errors.py
```diff
@@ -391,3 +391,26 @@ def test_repeated_kwarg():
 )
 def test_unparenthesized_genexp(source, no_errors):
     assert bool(_get_error_list(source)) ^ no_errors
+
+@pytest.mark.parametrize(
+    ('source', 'no_errors'), [
```
davidhalter, line 396: FYI: I usually try to not double invert things, because it's hard to reason about. For example here `no_errors=False` would mean that there are errors, whereas `has_errors=True` and `has_errors=False` is easier to understand. But this is not an API, so it just doesn't matter at all. Just a thing to think about (which you probably already do, but I thought I'd post it anyways).

PR #136, opened 2020-06-18
Ignore permission errors when saving to cache

davidhalter:
This might happen when a user doesn't have full access to his home directory.
Fixes davidhalter/jedi#1615

### DIFF
--- a/parso/_compatibility.py
+++ b/parso/_compatibility.py
@@ -51,10 +51,10 @@ except NameError:
     # Python 2.7 (both IOError + OSError)
     FileNotFoundError = EnvironmentError
 try:
-    # Python 2.7
+    # Python 3.3+
     PermissionError = PermissionError
 except NameError:
-    # Python 3.3+
+    # Python 2.7 (both IOError + OSError)
     PermissionError = EnvironmentError
 
 
--- a/parso/_compatibility.py
+++ b/parso/_compatibility.py
@@ -50,6 +50,12 @@ try:
 except NameError:
     # Python 2.7 (both IOError + OSError)
     FileNotFoundError = EnvironmentError
+try:
+    # Python 2.7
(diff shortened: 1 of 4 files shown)


test/test_cache.py
```diff
@@ -159,16 +162,29 @@ def test_inactive_cache(tmpdir, isolated_parso_cache):
...
+def test_permission_error(monkeypatch, recwarn):
+    def save(*args, **kwargs):
+        was_called[0] = True  # Python 2... Use nonlocal instead
+        raise PermissionError
+
+    was_called = [False]
+
+    monkeypatch.setattr(cache, '_save_to_file_system', save)
+    parse(path=__file__, cache=True, diff_cache=True)
+    assert was_called[0]
```
isidentical, line 190: Would you also add something to verify the warning has been raised. Something like `pytest.warns(Warning)` or `len(recwarn) > 0`

parso/_compatibility.py
```diff
@@ -50,6 +50,12 @@ def u(string):
 except NameError:
     # Python 2.7 (both IOError + OSError)
     FileNotFoundError = EnvironmentError
+try:
+    # Python 2.7
+    PermissionError = PermissionError
+except NameError:
+    # Python 3.3+
```
isidentical, line 57: I find this comments a bit confusing, shouldn't this be `Python 2.7` and the upper one `Python 3.3+`?

test/test_cache.py
davidhalter: Very good point. I even had `recwarn` in the function signature, but then forgot to implement it. Thanks for mentioning it!

parso/_compatibility.py
davidhalter: Yeah it's definitely wrong. No idea how I messed that up :)

PR #140, opened 2020-06-28
Fix #139: newlines in async for comprehension

### LINKED ISSUE #139: Syntax error in async for comprehension depending on newline
Hello,

Using the latest master of `parso`, with Python 3.7.6, in an async for comprehension, it reports a `SyntaxError` depending on whether or not there is a newline before `]` (in the case of a list comprehension).

This is accepted:
```python
async def go():
    return [1 
        async for _ in range(5)]
```

But this isn't even though CPython is fine with it:
```python
async def go():
    return [1 
        async for _ in range(5)
    ]
```

I first experienced it with Jedi and I've reproduced it in `parso` like this:
```python
import parso

script_success = """
async def go():
    return [1 async for _ in range(5)]
"""

script_success_2 = """
async def go():
    return [1 
        async for _ in range(5)]
"""

script_fail = """
async def go():
    return [1 
        async for _ in range(5)
    ]
"""

grammar = parso.load_grammar()
for module in map(grammar.parse, [script_success, script_success_2, script_fail]):
    errors = grammar.iter_errors(module)
    print(errors)
    for err in errors:
        print(err.message)

# Output:
# []
# []
# [<Issue: 901>, <Issue: 901>]
# SyntaxError: invalid syntax
# SyntaxError: invalid syntax
```

Thanks !

Kazy:
I decided to take a stab at this. I don't know if this is the right fix, and if the test is the right place (should it be in `test_tokenize.py` ?).

I traced back the difference between the parsing of a for comprehension and an async one to the tokenization. In a normal for-comprehension, the `]` token has `\n     ` as a prefix, while in the case of `async for` it has `NEWLINE` token which isn't part of the DFA transitions and is thus a `SyntaxError`.

Removing `async` from the `ALWAYS_BREAK_TOKENS` fixes this, all the tests are passing. I don't know if this can have consequences in other uses of `async`. Codes like this are still rejected:
```python
async
def foo():
   ....

# or
async
    def foo():
        ...
```

Now it also correctly parses all possible insertion of newlines in the async for-comprehension, just like a sync one.
```python
async def foo():
    [1 
    async
    for
    _
    in
    range(5)
    ]
```

### DIFF
--- a/test/test_pgen2.py
+++ b/test/test_pgen2.py
@@ -87,6 +87,39 @@ def test_async_for(works_ge_py35):
     works_ge_py35.parse("async def foo():\n async for a in b: pass")
 
 
+@pytest.mark.parametrize("body", [
+    """[1 async for a in b
+    ]""",
+    """[1 async
+    for a in b
+    ]""",
+    """[
+    1
+    async for a in b
+    ]""",
+    """[
+    1
+    async for a
+    in b
+    ]""",
+    """[
+    1
+    async
+    for
+    a
+    in
+    b
+    ]""",
+    """  [
+    1 async for a in b
+  ]""",
+])
+def test_async_for_comprehension_newline(works_ge_py36, body):
+    # Issue #139
+    works_ge_py36.parse("""async def foo():
+    {}""".format(body))
+
+
 def test_async_with(works_ge_py35):
     works_ge_py35.parse("async def foo():\n async with a: pass")
 
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -51,6 +51,7 @@ Simon Ruggier (@sruggier)
 Élie Gouzien (@ElieGouzien)
 Tim Gates (@timgates42) <tim.gates@iress.com>
 Batuhan Taskaya (@isidentical) <isidentical@gmail.com>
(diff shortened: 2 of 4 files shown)


test/test_pgen2.py
```diff
@@ -87,6 +87,17 @@ def test_async_for(works_ge_py35):
...
+def test_async_for_comprehension_newline():
+    # Issue #139
+    _parse("""
+async def foo():
+    [1 async 
+    for a in b 
+    ]
+""")
+
+
```
isidentical, line 100: It would be actually nicer to add something like `works_ge_py36` to `conftest.py` and then use that fixture.  Also please add some other tests with different versions such as these;
```py

async def foo():
    [1
    async
    for a in b 
    ]

async def foo():
    [1 async 
    for a in b 
    ]

async def foo():
    [
        1 async for a in b 
    ]
```
and possibly more.

davidhalter: :+1: for `works_ge_py36`.

Kazy: Updated !

Kazy:
> (for a better look on the test, if you want you can move the last triple quotes to the upper line (so it wont break the indentation))

Done !

PR #143, opened 2020-06-30
Parse alpha, beta and rc versions strings.

### LINKED ISSUE #142: Detect alpha, beta and rc versions of Python
```
In [1]: from parso.utils import parse_version_string

In [2]: parse_version_string('3.10.0a0')

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-70c767a0d0a3> in <module>
----> 1 parse_version_string('3.10.0a0')

~/dev/parso/parso/utils.py in parse_version_string(version)
    183         raise TypeError('version must be a string like "3.8"')
    184
--> 185     return _parse_version(version)

~/dev/parso/parso/utils.py in _parse_version(version)
    132     match = re.match(r'(\d+)(?:\.(\d{1,2})(?:\.\d+)?)?$', version)
    133     if match is None:
--> 134         raise ValueError('The given version is not in the right format. '
    135                          'Use something like "3.8" or "3".')
    136

ValueError: The given version is not in the right format. Use something like "3.8" or "3".
```

Carreau:
fixes #142

### DIFF
--- a/parso/utils.py
+++ b/parso/utils.py
@@ -129,7 +129,7 @@ def version_info():
 
 
 def _parse_version(version):
-    match = re.match(r'(\d+)(?:\.(\d{1,2})(?:\.\d+)?)?$', version)
+    match = re.match(r'(\d+)(?:\.(\d{1,2})(?:\.\d+)?)?((a|b|rc)\d)?$', version)
     if match is None:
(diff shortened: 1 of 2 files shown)


davidhalter:
Thanks!

PR #145, opened 2020-07-21
Let consumers know that we have type annotations

PeterJCLaw:
As well as the type stubs, this includes both the `py.typed` flag file (for tools) and the classifier (for people).

I've manually checked this by installing this against my development `jedi`. Not only does it make the `mypy` run faster, but it also solves a bunch of the errors I was getting. There's one remaining (lack of `parso.python.tree.search_ancestor`), though I'll look at that separately.


davidhalter:
Thanks!

PR #146, opened 2020-07-24
Dropping Python <3.6

davidhalter:
I'm currently trying to get rid of a lot of old code. The next release will not feature the old Pythons anymore :tada: !

IMO we should also drop Python 2 parsing. Do you agree @isidentical ? (Also don't review yet, a lot is going to change)

I guess after merging this branch anyone can feel free to make various Python 3 improvements.


isidentical:
> IMO we should also drop Python 2 parsing. Do you agree @isidentical ? (Also don't review yet, a lot is going to change)

I won't have argument against it unless you need that feature in jedi, since there is a lot of Python 2 code is lurking around out there. -

davidhalter:
Jedi is dropping Python 2 as well. The branch there is already finished. People can still pin the older versions of both of them.

isidentical:
> Jedi is dropping Python 2 as well. The branch there is already finished. People can still pin the older versions of both of them.

Awesome! So, there is no blockers for us to drop 2.

isidentical:
I won't be available in the next week. If you'd be OK, I plan to submit a prelimanary review on the tomorrow evening and then you're free to do final adjustments or other changes and land it to upstream?

davidhalter:
Yeah, that's definitely ok.

IMO you don't have to review everything. There are a ton of small changes in there. I'm of course happy if you do, but mostly there's one remaining question:

- I have formatted the whole code base in a way that works with flake8. I have also added an automatic run to travis that checks for this formatting. IMO this is a very low key tool (contrary to black). I like it, but I'm also happy to drop it, since I'm not a big fan of linters anyway.

I have also added a mypy run to the CI, that checks for typing issues, but I doubt that this is very controversial.

davidhalter:
Should be good now. There are a lot of semi related changes. It's just that I never really go through a big part of the code base. So I took the freedom to edit some of these things as well.

The biggest and only real change (apart from flake8/mypy) is that using `path.Path` is now used internally for paths.

davidhalter:
Also feel free to not review this. I think this is a huge pain to review and it seems like it will take a lot of time. The flake8 question is probably the biggest one.

conftest.py
```diff
@@ -30,7 +29,7 @@ def clean_parso_cache():
     """
     old = cache._default_cache_path
     tmp = tempfile.mkdtemp(prefix='parso-test-')
-    cache._default_cache_path = tmp
+    cache._default_cache_path = Path(str(tmp))
```
isidentical, line 32: Do we need this `str()` call here? AFAIK `tempfile.mkdtemp` always returns a string. If you wish to ensure, I guess `os.fspath` would be more convenient but as I said I am unsure whether needed or not.

docs/conf.py
```diff
@@ -273,7 +273,7 @@
 # -- Options for intersphinx module --------------------------------------------
 
 intersphinx_mapping = {
-    'http://docs.python.org/': ('https://docs.python.org/3.6', None),
+    'http://docs.python.org/': ('https://docs.python.org/3.8', None),
```
isidentical, line 276: Maybe just refer to `docs.python.org/3/` which would open the most recent stable version (3.8.5 in this case)

parso/cache.py
```diff
@@ -64,21 +58,19 @@
...
+It's a bit similar to `sys.implementation.cache_tag`.
+See: http://docs.python.org/3/library/sys.html#sys.implementation
 """
 
 
 def _get_default_cache_path():
     if platform.system().lower() == 'windows':
-        dir_ = os.path.join(os.getenv('LOCALAPPDATA')
-                            or os.path.expanduser('~'), 'Parso', 'Parso')
+        dir_ = Path(os.getenv('LOCALAPPDATA') or os.path.expanduser('~'), 'Parso', 'Parso')
```
isidentical, line 68: Why not just
```suggestion
        dir_ = Path(os.getenv('LOCALAPPDATA') or '~', 'Parso', 'Parso')
```

parso/grammar.py
```diff
@@ -1,9 +1,12 @@
...
 import os
+from typing import Generic, TypeVar, Union, Dict, Optional, Any
+from pathlib import Path
 
-from parso._compatibility import FileNotFoundError, is_pypy
+from parso._compatibility import is_pypy
 from parso.pgen2 import generate_grammar
-from parso.utils import split_lines, python_bytes_to_unicode, parse_version_string
+from parso.utils import split_lines, python_bytes_to_unicode, \
+    PythonVersionInfo, parse_version_string
```
isidentical, line 9: Is it a style preference for you to use backslashes instead of indented groups like
```
from x import (
    y,
    z,
    q
)
```

parso/grammar.py
```diff
@@ -216,45 +224,44 @@ def __init__(self, version_info, bnf_text):
...
+        return tokenize_lines(lines, version_info=self.version_info, **kwargs)
 
     def _tokenize(self, code):
         # Used by Jedi.
-        return tokenize(code, self.version_info)
+        return tokenize(code, version_info=self.version_info)
 
 
-def load_grammar(**kwargs):
+def load_grammar(*, language: str = 'python', version: str = None, path: str = None):
```
isidentical, line 234: I don't see the point of this `language` feature, at least for now. Would you mind briefly explaining it to me?

parso/python/pep8.py
```diff
@@ -679,6 +716,7 @@ def _analyse_non_prefix(self, leaf):
             for i, line in enumerate(leaf.value.splitlines()[1:]):
                 indentation = re.match(r'[ \t]*', line).group(0)
                 start_pos = leaf.line + i, len(indentation)
+                print(start_pos)
```
isidentical, line 719: Any reason to keep this `print`?

parso/tree.py
```diff
@@ -20,12 +18,12 @@ def search_ancestor(node, *node_types):
...
 
 
-class NodeOrLeaf(object):
+class NodeOrLeaf:
     """
     The base class for nodes and leaves.
     """
     __slots__ = ()
-    type = None
+    type: str
```
isidentical, line 26: Is there a code path where `type` is None? (I guess not if you said solely `str` for the annotation?)

davidhalter: No it shouldn't be.

parso/grammar.py
davidhalter: Absolutely. Removed.

parso/python/pep8.py
davidhalter: Fixed.

parso/grammar.py
davidhalter: I personally prefer backslashes, but I think it doesn't matter at all, so feel free to use indented groups as well.

docs/conf.py
davidhalter: Good point. Thanks!

conftest.py
davidhalter: Did not know this. Changed.

davidhalter:
Since I applied all of your proposed changes, I'm going to merge.

You didn't bring up the flake8 point, so I guess you are fine with it. Note that we can always remove it again if it's annoying or change to some other tool, etc.

isidentical:
I don't use flake8 in my personal projects but have no objections it being default here. If like find an annoying thing, we can just configure it to ignore that! I'm really happy to see that all these code refactored and thank you so much for making this happen 🎊 🥳

PR #152, opened 2020-09-24
Retrieve all kinds of assignment targets from with test

### LINKED ISSUE #151: with statement assigning to instance variable throws error
This code raises `ValueError: The name is not actually part of a with statement.` though it looks okay to me.

```
code = """
class Foo:
 def g(self):
    with open('file') as self.foo:
      pass
"""

import parso

grammar = parso.load_grammar()
module = grammar.parse(code)
g = module.children[0].children[-1].children[-1]  # points to Function g
withstmt = g.children[-1].children[-1]  # points to Withstmt

for name in withstmt.get_defined_names():
  # name.parent points to `PythonNode(trailer, [<Operator: .>, <Name: foo@4,30>])` instead of `withstmt`
  withstmt.get_test_node_from_name(name)
```

isidentical:
Resolves #151

### DIFF
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -775,8 +775,8 @@ class WithStmt(Flow):
         return names
 
     def get_test_node_from_name(self, name):
-        node = name.parent
-        if node.type != 'with_item':
+        node = search_ancestor(name, "with_item")
+        if node is None:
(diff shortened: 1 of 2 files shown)


davidhalter:
Makes sense, thanks!

PR #156, opened 2020-11-18
Properly check for invalid conversion character with f-string debugging syntax

gousaiyang:
When an f-string contains `=` debugging syntax in Python 3.8, the conversion character part may not be located at the position of `fstring_expr.children[2]`, thus its validity (only `!s`, `!r`, `!a` allowed) may not be properly checked. This PR will fix it.


isidentical (approved):
LGTM

isidentical:
Thanks @gousaiyang

Issue #158, opened 2020-11-18
Parso fails to parse some "unparenthesized" syntactic structures in f-string expression part


gousaiyang:
Currently `parso` cannot parse the following unparenthesized structures in f-string expression part, which are actually syntactically valid Python code:

- Unparenthesized iterable unpacking in expression lists: `f'{*x,}'`, `f'{*x, *y}'`, `f'{x, *y}'`, `f'{*x, y}'`
- Unparenthesized generator expression: `f'{x for x in [1]}'`
- Unparenthesized yield expression: `def foo(): f'{yield 1}'` (I just saw #157 is already opened for this)

isidentical:
Would you like to submit a patch?

gousaiyang:
Looks like updating the grammar is sufficient to solve the problem, I've created PR #159.

davidhalter:
Thanks @gousaiyang and @isidentical.

PR #159, opened 2020-11-19
Allow some unparenthesized syntactic structures in f-string expression part

### LINKED ISSUE #157: Unparenthesized yield is not allowed in the f-strings
```
>>> _get_error_list("def foo(): return f'{yield}'")
[<Issue: 901>]
```

gousaiyang:
Fixes #157, #158

### DIFF
--- a/parso/python/grammar36.txt
+++ b/parso/python/grammar36.txt
@@ -154,5 +154,5 @@ strings: (STRING | fstring)+
 fstring: FSTRING_START fstring_content* FSTRING_END
 fstring_content: FSTRING_STRING | fstring_expr
 fstring_conversion: '!' NAME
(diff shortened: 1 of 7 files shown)


parso/python/grammar36.txt
```diff
@@ -154,5 +154,5 @@ strings: (STRING | fstring)+
 fstring: FSTRING_START fstring_content* FSTRING_END
 fstring_content: FSTRING_STRING | fstring_expr
 fstring_conversion: '!' NAME
-fstring_expr: '{' testlist_comp [ fstring_conversion ] [ fstring_format_spec ] '}'
```
isidentical, line 157: Huh, I wonder why this is different than 3.7+

isidentical (approved):

isidentical:
Thanks @gousaiyang

PR #160, opened 2020-11-21
Support named unicode characters in f-strings

### LINKED ISSUE #154: Named unicode escapes with spaces marked as syntax error in f-strings
Given the following code:

```python
def bullet(text: str) -> str:
    return f'\N{BULLET} {text}'


def error(text: str) -> str:
    return f'\N{NO ENTRY} {text}'
```

A syntax error is marked for `ENTRY` but it is valid Python code.

thatch:
Fixes #154

The previous behavior misinterpreted the curly braces as enclosing an
expression.  This change does some cursory validation so we can still
get parse errors in the most egregious cases, but does not validate that
the names are actually valid, only that they are name-shaped and have a
chance of being valid.

The character names appear to obey a few rules:
* Case insensitive
* Name characters are `[A-Z0-9 \-]`
* Whitespace before or after is not allowed
* Whitespace in the middle may only be a single space between words
* Dashes may occur at the start or middle of a word

```py
f"\N{A B}"           # might be legal
f"\N{a b}"           # equivalent to above
f"\N{A     B}"       # no way
f"\N{    A B     }"  # no way
f"""\N{A
B}"""                # no way
```

For confirming this regex matches all (current) unicode character names:

```py
import re
import sys
import unicodedata

R = re.compile(r"[A-Za-z0-9\-]+(?: [A-Za-z0-9\-]+)*")

for i in range(sys.maxunicode):
    try:
        name = unicodedata.name(chr(i))
    except ValueError:
        # Some small values like 0 and 1 have no name, /shrug
        continue
    m = R.fullmatch(name)
    if m is None:
        print("FAIL", repr(name))
```

### DIFF
--- a/parso/python/tokenize.py
+++ b/parso/python/tokenize.py
@@ -110,9 +110,14 @@ def _get_token_collection(version_info):
             _create_token_collection(version_info)
         return result
 
-
-fstring_string_single_line = _compile(r'(?:\{\{|\}\}|\\(?:\r\n?|\n)|[^{}\r\n])+')
-fstring_string_multi_line = _compile(r'(?:[^{}]+|\{\{|\}\})+')
+unicode_character_name = r'[A-Za-z0-9\-]+(?: [A-Za-z0-9\-]+)*'
+fstring_string_single_line = _compile(
+    r'(?:\{\{|\}\}|\\N\{' + unicode_character_name +
+    r'\}|\\(?:\r\n?|\n)|\\[^\r\nN]|[^{}\r\n\\])+'
+)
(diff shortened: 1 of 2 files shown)


isidentical (approved):
LGTM. Could you please add a few tests where these unicode escapes used in the middle and at the end of the f-string. Like; `f"some {stuff} and \N{escape}"` etc.

thatch:
Additional tests in latest push.

isidentical:
Thanks for the patch @thatch! Also loved your work on https://github.com/thatch/python-grammar-changes

PR #162, opened 2020-11-27
Allow unparenthesized walrus in set literals, set comprehensions and indexes

gousaiyang:
This incorporates two recent changes in CPython upstream:

- As per python/cpython#23332, unparenthesized assignment expressions are allowed in set literals and set comprehensions (but not in dict literals and dict comprehensions). This change applies to Python 3.9+. Examples: `{x := 1, 2, 3}`, `{x4 := x ** 5 for x in range(7)}`.
- As per python/cpython#23317, unparenthesized assignment expressions are allowed in indexes (but not in slices). This change applies to Python 3.10+. Examples: `a[b:=0]`, `a[b:=0, c:=0]`.

As for the code change, I cannot directly use `namedexpr_test` in `dictorsetmaker` and `subscript` because that will produce ambiguous rules.

### DIFF
--- a/parso/python/grammar310.txt
+++ b/parso/python/grammar310.txt
@@ -124,14 +124,14 @@ atom: ('(' [yield_expr|testlist_comp] ')' |
 testlist_comp: (namedexpr_test|star_expr) ( comp_for | (',' (namedexpr_test|star_expr))* [','] )
 trailer: '(' [arglist] ')' | '[' subscriptlist ']' | '.' NAME
 subscriptlist: subscript (',' subscript)* [',']
-subscript: test | [test] ':' [test] [sliceop]
+subscript: test [':=' test] | [test] ':' [test] [sliceop]
 sliceop: ':' [test]
 exprlist: (expr|star_expr) (',' (expr|star_expr))* [',']
 testlist: test (',' test)* [',']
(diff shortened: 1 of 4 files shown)


isidentical (approved):

isidentical:
I believe it looks fine, thanks @gousaiyang! @davidhalter can you check this out? "The ambiguity arises in"

davidhalter:
Sorry for the delay (was awaiting my firstborn :)), but thanks for the change!

I think this is definitely fine, but I will check why this causes ambiguities.

IMO we will have to find a (PEG) solution for the parser anyway during the next year, so this grammar change is not a problem anyway.

isidentical:
> Sorry for the delay (was awaiting my firstborn :)

🚀 Congratulations!

davidhalter:
I just tried to check why the ambiguity problem arises:

```
subscript: test [':=' test] | [test] ':' [test] [sliceop]
vs.
subscript: namedexpr_test | [test] ':' [test] [sliceop]
```

The problem in the latter example is that for all the first tokens (e.g. for a `lambda` token), the LL parser does not know if it should use `test` or `namedexpr_test`, because there's an alternative `[test]`.

So this is definitely a limitation of our current parser.

davidhalter:
Also released 0.8.1 with this.

PR #163, opened 2020-11-30
Travis-ci: added support for ppc64le

dthadi3:
Added power support for the travis.yml file with ppc64le. This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 dist: xenial
+arch:
+  - AMD64
+  - ppc64le
 language: python
 python:
   - 3.6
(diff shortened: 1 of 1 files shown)


isidentical:
Is this an automated commit? If not, I wonder why do we need to test in ppc64le, as far as I can tell there is no arch-specific code in the parso.

dthadi3:
Thanks for your response @isidentical it's not an automated commit FYI, @gerrith3 can you please respond for the query.

dthadi3:
Thanks for your reply @isidentical It's not the automated commit. @gerrith3 can you please respond to the query.

davidhalter:
@isidentical Please don't merge this without my approval. I really don't see why we need this.

isidentical:
> @isidentical Please don't merge this without my approval. I really don't see why we need this.

I also agree.

gerrith3:
Hi @isidentical , @davidhalter it is always hard to tell from the outside wihch code is truly architecture independent.  Python does have plenty of room for arch dependencies and without evaluating the project, checking with the maintainers about current and future intentions, it is very hard to tell if all contributors will continue to keep a project arch independent.  The advantage of testing on Power is that it will immediately flag any deviation from your intention to be archtecture independent - if the ppc64le build ever fails when the x86-64 bulid succeeds, it would be obvious that someone committed a change that was not arch neutral.  It is up to you - we are just trying to maintain currency of top of tree for more projects.  If you feel there is no chance of a regression/arch dependent code, then you are right, you probably don't need this test.  However, if you want some automation to help ensure that the tree stays arch independent, this is an available option.

davidhalter:
@gerrith3 Thanks for the explanation. I highly doubt that a Python parser will ever introduce arch dependencies as long as it stays pure Python.

There is indeed a chance that parso gains a Rust backend to improve the speed, but at that point we will have to see which architectures we support. We might get back to you if we want to support Power.

At the moment I doubt that the Rust change will ever make it into parso, it will probably be a separate project.

PR #168, opened 2020-12-30
Fix various issues regarding starred expressions

### LINKED ISSUE #166: f2format use case
apparently f2format is able to raise a bug inside parso :
https://github.com/pybpc/f2format/issues/16

gousaiyang:
- Properly check for starred expression deletion
- Check for starred expressions not in tuple/list/set (when not in assignment)
- Fix a bug that considered starred expression assignment `[*x] = 1` as invalid
- Parenthesized single starred expression `(*x)` is no longer valid anywhere in Python 3.9+. Therefore `a, (*b), c = [1, 2, 3]` will be invalid in Python 3.9+. See [bpo-40631](https://bugs.python.org/issue40631).
- Enhance test cases for valid and invalid `del` statements and starred expressions

Fixes #166

### DIFF
--- a/parso/python/errors.py
+++ b/parso/python/errors.py
@@ -753,6 +753,16 @@ class _StarExprRule(SyntaxRule):
                     return False
             return False
 
+        if self._normalizer.version >= (3, 9):
+            ancestor = node.parent
+        else:
+            ancestor = _skip_parens_bottom_up(node)
+        # starred expression not in tuple/list/set
+        if ancestor.type not in (*_STAR_EXPR_PARENTS, 'dictorsetmaker') \
+                and not (ancestor.type == 'atom' and ancestor.children[0] != '('):
+            self.add_issue(node, message="can't use starred expression here")
+            return
+
         if check_delete_starred(node):
             if self._normalizer.version >= (3, 9):
                 self.add_issue(node, message="cannot delete starred")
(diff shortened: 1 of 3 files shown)


davidhalter:
@isidentical, if you have time please review. I think you know this code better than me. If you don't have time, I'm happy to do it.

parso/python/errors.py
```diff
@@ -730,9 +742,34 @@ def is_issue(self, node):
...
 
     def is_issue(self, node):
+        def check_delete_starred(node):
+            while node.parent is not None:
+                node = node.parent
+                if node.type == 'del_stmt':
+                    return True
+                if node.type not in (*_STAR_EXPR_PARENTS, 'atom'):
+                    return False
+            return False
```
isidentical, line 754: You don't have to re-implement parent searching system here, I think you reduce this function (and your usage to) just this;
```
nearest_ancestor = search_ancestor(node, 'del_stmt', 'atom', *_STAR_EXPR_PARENTS)
if nearest_ancestor == 'del_stmt':
    ...
```

gousaiyang: `search_ancestor` locates the nearest ancestor of one of the given types, but here I need to find the nearest ancestor which is **NOT** of the given types (`(*_STAR_EXPR_PARENTS, 'atom')`).

isidentical: Ah, I see.

isidentical (approved):
It LGTM in general!

davidhalter:
Hi @gousaiyang 

Due to your recent contributions, I have added you as a collaborator. Thanks a lot for everything you did. I hope you don't stop :) I have talked to @isidentical about this and we both think it would be great to have you onboard.

In general this shouldn't change too much. All of us still do pull requests, so it would be nice if you didn't push bigger changes directly. IMO it's fine to push small bugfixes by yourself (as long as there is a test).

If you want the ability to do parso releases, I'm happy to give you the PyPI access.

Happy New Year!
~ Dave

@isidentical Generally feel free to merge PRs if you think they are fine. You don't need to wait for my approval.

davidhalter:
@gousaiyang Please do email me (davidhalter88@gmail.com) so that I have a way to reach you if there's ever something we want to discuss "internally".

PR #169, opened 2021-01-03
Add type annotations to start- and end-pos attributes

PeterJCLaw:
These are frequently used within consuming code, so having annotations avoids others needing to work around mypy errors from them.

I've added a small number of other annotations where it made sense to do so, though I've tried to keep this PR limited to just the position attributes.

### DIFF
--- a/parso/tree.py
+++ b/parso/tree.py
@@ -192,7 +192,7 @@ class Leaf(NodeOrLeaf):
         :py:func:`str` Typically a mixture of whitespace and comments. Stuff
         that is syntactically irrelevant for the syntax tree.
         '''
-        self.parent = None  # type: Optional[BaseNode]
+        self.parent: Optional[BaseNode] = None
         '''
         The parent :class:`BaseNode` of this leaf.
         '''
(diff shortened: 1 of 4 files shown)


parso/tree.py
```diff
@@ -191,17 +192,17 @@ def __init__(self, value, start_pos, prefix=''):
         :py:func:`str` Typically a mixture of whitespace and comments. Stuff
         that is syntactically irrelevant for the syntax tree.
         '''
-        self.parent = None
+        self.parent = None  # type: Optional[BaseNode]
```
davidhalter, line 195: We should probably use variable annotations. Python < 3.6 is not supported anymore.

parso/tree.py
```diff
@@ -258,26 +259,26 @@ class BaseNode(NodeOrLeaf):
...
     __slots__ = ('children', 'parent')
 
-    def __init__(self, children):
+    def __init__(self, children: List[NodeOrLeaf]) -> None:
         self.children = children
         """
         A list of :class:`NodeOrLeaf` child nodes.
         """
-        self.parent = None
+        self.parent = None  # type: Optional[BaseNode]
```
davidhalter, line 267: Same here.

parso/tree.py
PeterJCLaw: Aha, I'd forgotten that was gone now. 1139e53

parso/tree.py
PeterJCLaw: 1139e53

davidhalter:
Thanks!

PR #170, opened 2021-01-26
Show path in warning message

pjvandehaar:
I get the warning message "Tried to save a file to %s, but got permission denied." in my ipython repl and it shows "%s" instead of a path.  This commit fixes that.


davidhalter:
Thanks for the fix!

PR #173, opened 2021-02-16
Only apply non-default params logic prior to the star (fixes #161)

### LINKED ISSUE #161: SyntaxError: non-default argument follows default argument
For keyword-only parameters, non-defaulted can in fact follow a default argument. 
```
>>> def foo(*, bar=100, baz): pass
... 
>>> [error] = _get_error_list("def foo(*, bar=100, baz): pass")
>>> error.message
'SyntaxError: non-default argument follows default argument'
```

bryanforbes:


gousaiyang (commented):
LGTM

davidhalter:
@gousaiyang Feel free to merge if you think it is good enough :)

If we think the merge was a mistake we can still revert (worst case).

PR #177, opened 2021-02-24
Fix flaky test_cache_last_used_update

bobrik:
Sometimes time moves slowly and strict comparison is not enough:

```
>       assert now < node_cache_item.last_used < time.time()
E       assert 1614147958.869299 < 1614147958.869299
E        +  where 1614147958.869299 = <parso.cache._NodeCacheItem object at 0x10456fe80>.last_used
E        +  and   1614147958.869299 = <built-in function time>()
E        +    where <built-in function time> = time.time

test/test_cache.py:149: AssertionError
```

In particular, macOS timings can be a bit coarse.

The test failure is from Apple Silicon M1.


davidhalter:
Thanks!

PR #178, opened 2021-02-25
Added .editorconfig according to the style used in this project.

KOLANICH:


isidentical:
Same Q with your previous PR, what is the motivation (please use PR description)?

KOLANICH:
This file is a musthave for any project dealing with source code because it solves some painful and highly hollywarable problems.

I mean the following.

In python ecosystem there exists a document called PEP 8. This "standard" recommends some one of the worst choices, like using spaces for indentation (the worst one in this case). This PEP is about python standard library, but people writing python libs tend to think that any python lib must use it.

Of course not all the people aggree to follow these controversal practices, instead they prefer to follow better ones. But if a person has to contribute to a project, he has to follow the practice used in the project.

Of course, editors allow to set their global settings to the ones used in projects. But it is insane to do that!:

* Editors allow multiple files of multiple projects be opener concurently; each project has own requirements, the result is that a user cannot use this feature.
* It is an additional burden to a user to go into settings each time he works on projects with different requirements.

Editors having concept of projects may have own configurations, but it is an additional burden on contributors to require them to set up it for each project manually.

One can store project files for each editor ... but it would be unwise, and also won't solve the problem with editors without a concept of a project.

The real solution are editorconfig files. They are not tied to any specific editor. They allow to achieve an uniform view by setting tab size to a specific value for all the editors.

In other words they allow devs not to write code instead of taking part in hollywars like tabs vs spaces.

So, basically, if a project doesn't have an editorconfig, persons contributing to many projects have to create one (or copy from som other project with similar conventions). It is an additional burden. It is better to have this file already in a repo.

isidentical:
We've never had issues with this file, and the PR seems pre-emptive. Though I'll leave this to @davidhalter

davidhalter:
> This "standard" recommends some one of the worst choices, like using spaces for indentation (the worst one in this case). 

There are reasons for both. I really think you don't understand why some people prefer one or the other. Essentially it doesn't matter at all. It is not something that you should be even thinking about. You should care about datastructures, APIs and security. Most other things don't really matter that much. Arguing about whitespace and linting rules is really just a waste of time.

> This PEP is about python standard library, but people writing python libs tend to think that any python lib must use it.

Because that's how it is for most languages. And that's a big advantage. It's just the same in most projects. When I use Go I use tabs and when I use Rust I use 4 spaces, because that's the "standard". 

> not all the people aggree to follow these controversal practices

In that case the burden is on them I guess ;-).

PR #184, opened 2021-04-07
Unparenthesized lambda no longer allowed in comp_if since Python 3.9

gousaiyang:
Since Python 3.9, unparenthesized lambda expressions (specifically, `lambdef_nocond`) are no longer allowed in `comp_if`. Thus, the grammar `comp_if: 'if' test_nocond [comp_iter]` can be simplified to be `comp_if: 'if' or_test [comp_iter]`, and then `test_nocond` and `lambdef_nocond` can be removed. This behavior was already there ever since 3.9.0, although the documentation was only updated recently (in [bpo-43755](https://bugs.python.org/issue43755)) to reflect this.

### DIFF
--- a/parso/python/grammar310.txt
+++ b/parso/python/grammar310.txt
@@ -97,9 +97,7 @@ suite: simple_stmt | NEWLINE INDENT stmt+ DEDENT
 
 namedexpr_test: test [':=' test]
 test: or_test ['if' or_test 'else' test] | lambdef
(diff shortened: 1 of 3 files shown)


davidhalter:
Thanks!

davidhalter:
Deleted your branch as well. Let me know if that's not OK and I won't do it in the future!

gousaiyang:
It's great. I always remove PR branches after merging.

Issue #155, opened 2020-11-03
Travis -> Github Actions


isidentical:
It would be nicer for us to move from travis to github actions, since the GA ecosystem is much bigger now (which we can automate a few tasks, like publishing to PyPI) and travis's new open-source pricing model (not sure whether if effects us or not though).

davidhalter:
I think the Travis pricing model change does not affect us, but I'm not 100% sure. If you want to change to GitHub Actions, I'm totally fine with that.

About publishing to PyPI, I'm not sure I want that. I actually like that a cloud service cannot do that by accident (or because of a security issue in GitHub Actions). If you want the necessary rights to do releases, I'm happy to give you access to publish on PyPI. It's just that I have my concerns about the security/safety aspect.

The releases are so rare that I doubt that we need to change something there. However as a CI I have no problem at all with this.

isidentical:
> About publishing to PyPI, I'm not sure I want that. I actually like that a cloud service cannot do that by accident (or because of a security issue in GitHub Actions). If you want the necessary rights to do releases, I'm happy to give you access to publish on PyPI. It's just that I have my concerns about the security/safety aspect.

Oh, okay. Was just giving an example of automations.

blueyed:
FWIW deploying to PyPI is also possible from Travis CI.

gousaiyang:
I've created #185 for this. Travis CI (`travis-ci.com`) stopped running on my projects several months ago. When you have used up the "initial credits", you either have to become a paid customer, or email them to apply for credits for open source projects. Today when I see the [parso Travis CI page](https://travis-ci.org/github/davidhalter/parso) (which is on `travis-ci.org`), it says `Please be aware travis-ci.org will be shutting down in several weeks, with all accounts migrating to travis-ci.com. Please stay tuned here for more information.`. Probably this is the time to move to GitHub Actions.

My current PR tries to preserve the original build jobs. Whether to use CI for automated publishing is our freedom of choice and is a separate thing to consider in the future if we want to do that.

PR #185, opened 2021-04-07
Travis CI -> GitHub Actions

gousaiyang:
Move from Travis CI to GitHub Actions (#155)


davidhalter:
Just FYI I did the port for Jedi: github.com/davidhalter/jedi

Since it uses similar ideas you could probably find some inspiration there if some tests won't pass (though 3.10 is probably a different kind of beast :)).

gousaiyang:
`coveralls` is working now. For 3.10, looks like either `parso` or `pytest` does not fully support 3.10 as for now. For Travis CI we had a config `allow_failures` to allow `nightly` to fail without failing the whole build. For GitHub Actions I'm using a similar thing `continue-on-error` for `3.10-dev`. The workflow is actually considered to be successful when we look at [the Actions tab](https://github.com/davidhalter/parso/actions), since all checks except `3.10-dev` passed, and `3.10-dev` is allowed to fail. But we may still see red cross marks ❌ displayed in some places, and it won't say "all checks passed". Maybe this is the current user experience of GitHub Actions `continue-on-error`.

davidhalter:
Shouldn't we just remove support for 3.10 for now? Parso does not support 3.10 parsing in a meaningful way anyway and I'm pretty sure there are going to be a ton of issues because of the parser changes. So I'd rather start debugging once an actual rc is out.

gousaiyang:
I've disabled `3.10-dev` build. Typically "experimental" builds are intended to discover incompatibility with `nightly` at an earlier time. But if the red cross mark is too annoying for us, it's totally fine to disable them for now.

davidhalter:
Thanks a lot for this!

> But if the red cross mark is too annoying for us, it's totally fine to disable them for now.

I think that for now this is the better choice. I agree that we should probably work on 3.10 support though!

PR #186, opened 2021-04-22
Relax a test regex to match new enum repr in Python 3.10.0a7+

hroncok:
bpo-40066: Enum: adjust repr() to show only enum and member name (not value,
nor angle brackets) and str() to show only member name.
https://bugs.python.org/issue40066


gousaiyang (approved):
Thanks!

Issue #113, opened 2020-05-22
Dump parso parse tree


isidentical:
Is there a way to dump the parse tree of parso, just like `ast.dump`?

davidhalter:
At the moment there is just pickle. There's no other format for now (and I'm not sure there should be, because it would basically be all slow).

isidentical:
> At the moment there is just pickle. There's no other format for now (and I'm not sure there should be, because it would basically be all slow).

I meant dumping as in printing the screen a visual tree, just like [this](https://github.com/python/cpython/blob/master/Tools/peg_generator/pegen/grammar_visualizer.py).

davidhalter:
Ah, ok. I never really thought about this or needed it.

I would be fine adding it, we would just need to come up with a good formatting and a concise method name.

gousaiyang:
Implemented in #187.

PR #187, opened 2021-05-03
Add function to dump parser tree

gousaiyang:
Add `parso.tree.dump` function to dump the parser tree (implement #113). The design mimics [`ast.dump`](https://docs.python.org/3/library/ast.html#ast.dump). Specifying `indent=4` will produce a readable format, while `indent=None` will produce a single-line compact representation. The produced dump can be `eval`'d back into a valid parso node object, which may be helpful for debugging in interactive REPL.

A sample dump:

```python
>>> import parso
>>> module = parso.parse('lambda x, y: x + y')
>>> print(repr(module))  # cannot see full children information via repr()
<Module: @1-1>
>>> print(parso.tree.dump(module, indent=4))  # full information dumped
Module([
    Lambda([
        Keyword('lambda', (1, 0)),
        Param([
            Name('x', (1, 7), prefix=' '),
            Operator(',', (1, 8)),
        ]),
        Param([
            Name('y', (1, 10), prefix=' '),
        ]),
        Operator(':', (1, 11)),
        PythonNode('arith_expr', [
            Name('x', (1, 13), prefix=' '),
            Operator('+', (1, 15), prefix=' '),
            Name('y', (1, 17), prefix=' '),
        ]),
    ]),
    EndMarker('', (1, 18)),
])
```

A number of small changes are made to make the dump fully "round-trip" (i.e. `eval(dump(node))` recovers the `node`):
- `Function` and `Lambda` node constructors are slightly modified to accept a `children` list which already contains `Param` objects (so that we don't need to process them with `_create_params` again).
- The `parent` argument of the `Param` node constructor becomes optional (to make generating round-trippable dumps easier).
- In `BaseNode` constructor, set all children's parent to `self` (to fix the parent relationship when we construct a new parse tree from the dump).

Question: Should we export this function just as `parso.tree.dump` or even `parso.dump`? And what about exporting `parso.tree.search_ancestor` as `parso.search_ancestor`?

### DIFF
--- a/parso/python/prefix.py
+++ b/parso/python/prefix.py
@@ -40,6 +40,14 @@ class PrefixPart:
             self.start_pos
         )
 
+    def search_ancestor(self, *node_types):
+        node = self.parent
+        while node is not None:
+            if node.type in node_types:
+                return node
+            node = node.parent
+        return None
+
 
 _comment = r'#[^\n\r\f]*'
 _backslash = r'\\\r?\n'
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -49,8 +49,7 @@ except ImportError:
     from collections import Mapping
 from typing import Tuple
 
-from parso.tree import Node, BaseNode, Leaf, ErrorNode, ErrorLeaf, \
-    search_ancestor
+from parso.tree import Node, BaseNode, Leaf, ErrorNode, ErrorLeaf
 from parso.python.prefix import split_prefix
 from parso.utils import split_lines
 
@@ -796,7 +795,7 @@ class WithStmt(Flow):
         return names
 
     def get_test_node_from_name(self, name):
-        node = search_ancestor(name, "with_item")
+        node = name.search_ancestor("with_item")
         if node is None:
             raise ValueError('The name is not actually part of a with statement.')
         return node.children[0]
@@ -1191,7 +1190,7 @@ class Param(PythonBaseNode):
         """
         Returns the function/lambda of a parameter.
         """
-        return search_ancestor(self, 'funcdef', 'lambdef')
+        return self.search_ancestor('funcdef', 'lambdef')
 
     def get_code(self, include_prefix=True, include_comma=True):
         """
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -549,13 +549,9 @@ class Function(ClassOrFunc):
         super().__init__(children)
         parameters = self.children[2]  # After `def foo`
         parameters_children = parameters.children[1:-1]
-        input_has_param = False
-        for child in parameters_children:
-            if isinstance(child, Param):
-                input_has_param = True
         # If input parameters list already has Param objects, keep it as is;
         # otherwise, convert it to a list of Param objects.
-        if not input_has_param:
+        if not any(isinstance(child, Param) for child in parameters_children):
             parameters.children[1:-1] = _create_params(parameters, parameters_children)
 
     def _get_param_nodes(self):
@@ -660,13 +656,9 @@ class Lambda(Function):
         super(Function, self).__init__(children)
         # Everything between `lambda` and the `:` operator is a parameter.
         parameters_children = self.children[1:-2]
-        input_has_param = False
-        for child in parameters_children:
-            if isinstance(child, Param):
-                input_has_param = True
         # If input children list already has Param objects, keep it as is;
         # otherwise, convert it to a list of Param objects.
-        if not input_has_param:
+        if not any(isinstance(child, Param) for child in parameters_children):
             self.children[1:-2] = _create_params(self, parameters_children)
 
     @property
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -553,8 +553,6 @@ class Function(ClassOrFunc):
         for child in parameters_children:
             if isinstance(child, Param):
                 input_has_param = True
-                # Fix parent relationship of Param children.
-                child.parent = parameters
         # If input parameters list already has Param objects, keep it as is;
         # otherwise, convert it to a list of Param objects.
         if not input_has_param:
@@ -666,8 +664,6 @@ class Lambda(Function):
         for child in parameters_children:
             if isinstance(child, Param):
                 input_has_param = True
-                # Fix parent relationship of Param children.
-                child.parent = self
         # If input children list already has Param objects, keep it as is;
         # otherwise, convert it to a list of Param objects.
         if not input_has_param:
@@ -1102,8 +1098,6 @@ class Param(PythonBaseNode):
     def __init__(self, children, parent=None):
         super().__init__(children)
         self.parent = parent
-        for child in children:
-            child.parent = self
 
     @property
     def star_count(self):
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -1100,7 +1100,7 @@ class Param(PythonBaseNode):
     """
     type = 'param'
 
-    def __init__(self, children, parent):
+    def __init__(self, children, parent=None):
         super().__init__(children)
         self.parent = parent
         for child in children:
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -549,7 +549,17 @@ class Function(ClassOrFunc):
     def __init__(self, children):
         super().__init__(children)
         parameters = self.children[2]  # After `def foo`
-        parameters.children[1:-1] = _create_params(parameters, parameters.children[1:-1])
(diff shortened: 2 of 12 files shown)


davidhalter:
>      Question: Should we export this function just as parso.tree.dump or even parso.dump? And what about exporting parso.tree.search_ancestor as parso.search_ancestor?

In general I would just put the dump function onto LeafOrNode (that might even eliminate some isinstance calls, though I'm not sure if that's practical, haven't reviewed the code in detail yet).

`search_ancestor` was definitely a mistake of mine. I would *definitely* put it onto `LeafOrNode` if I wasn't lazy. IMO we could just put it there and deprecate the other usage. What do you think?

isidentical:
I haven't had a chance to check the PR out, but I think the output is pretty cool!

gousaiyang:
I think it's a good idea to make `search_ancestor` and `dump` to be methods of `NodeOrLeaf`. That's more coherent with the rest of the API. I've made the change on this PR and also marked `parso.tree.search_ancestor` as deprecated.

parso/tree.py
```diff
@@ -1,22 +1,23 @@
...
     :param node: The ancestors of this node will be checked.
     :param node_types: type names that are searched for.
-    :type node_types: tuple of str
     """
-    while True:
-        node = node.parent
-        if node is None or node.type in node_types:
-            return node
+    warnings.warn("parso.tree.search_ancestor() is deprecated, "
+                  "use NodeOrLeaf.search_ancestor() instead", DeprecationWarning)
```
davidhalter, line 19: Can we maybe not have this warning yet? I know that a lot of people will be annoyed by it in IPython (because of some logger shenanigans) once we release the next parso version.

In general I feel like this is a good thing, but we should probably wait until Jedi uses the newer version. IMO there's no reason to hurry.

test/test_parser_tree.py
```diff
@@ -239,3 +240,29 @@ def test_with_stmt_get_test_node_from_name():
...
+        (sample_leaf, ('file_input',), sample_module),
+        (sample_leaf, ('file_input', 'arith_expr'), sample_node),
+        (sample_leaf, ('shift_expr',), None),
+        (sample_leaf, ('name', 'shift_expr',), None),
+        (sample_leaf, (), None),
+    ]
+)
+def test_search_ancestor(node, node_types, expected_ancestor):
+    assert node.search_ancestor(*node_types) is expected_ancestor
+    with pytest.deprecated_call():
```
davidhalter, line 266: On interesting! Didn't know about this, thanks!

parso/tree.py
```diff
@@ -266,9 +375,11 @@ def __init__(self, children: List[NodeOrLeaf]) -> None:
         """
         self.parent: Optional[BaseNode] = None
         '''
-        The parent :class:`BaseNode` of this leaf.
+        The parent :class:`BaseNode` of this node.
         None if this is the root node.
         '''
+        for child in children:
+            child.parent = self
```
davidhalter, line 382: I don't like this, because this is already done by the parser and would therefore be done twice. The problem is that this is done a lot and it would probably slow down the parser for no good reason :).

If we want this (which is fine) we should probably remove it in the parser so it is not done twice. If you run into bugs it's probably the diff parser. That thing is a beast and if you run into problems I'm happy to help you there.

parso/tree.py
```diff
@@ -173,13 +179,116 @@ def get_code(self, include_prefix=True):
...
+        :param node_types: type names that are searched for.
+        """
+        node = self.parent
+        while node is not None:
+            if node.type in node_types:
+                return node
+            node = node.parent
+        return None
+
+    def dump(self, *, indent: Optional[Union[int, str]] = None) -> str:
```
davidhalter, line 197: Shouldn't the default be something like 4? I don't think anybody wants to read a dump with the default `None` :D

But I'm fine with `None` as a default as well. Was just wondering why you picked it.

parso/python/prefix.py
```diff
@@ -40,6 +40,14 @@ def __repr__(self):
             self.start_pos
         )
 
+    def search_ancestor(self, *node_types):
+        node = self.parent
+        while node is not None:
+            if node.type in node_types:
+                return node
+            node = node.parent
+        return None
```
davidhalter, line 49: Interesting find! I would have forgotten about this. IMO it's not necessary here and we can probably remove this function. What do you think?

test/test_dump_tree.py
```diff
@@ -0,0 +1,165 @@
...
+            ])''')),
+    ]
+)
+def test_dump_parser_tree(indent, expected_dump):
+    code = "lambda x, y: x + y"
+    module = parse(code)
+    assert module.dump(indent=indent) == expected_dump
+
+    # Check that dumped tree can be eval'd to recover the parser tree and original code.
+    recovered_code = eval(expected_dump).get_code()
```
davidhalter, line 101: This is actually pretty sweet!

davidhalter (commented):
This is definitely a good patch. There are a few issues that I have, but most of them are minor. I think there's only one thing that matters to me: The change in `Node.__init__`.

Thanks!

parso/python/prefix.py
gousaiyang: [This place](https://github.com/davidhalter/parso/blob/f2b1ff942933a43c34d93e3359e7115d20cc2ee3/parso/python/pep8.py#L731) and [this place](https://github.com/davidhalter/parso/blob/f2b1ff942933a43c34d93e3359e7115d20cc2ee3/parso/python/pep8.py#L127) are calling the old `parso.tree.search_ancestor` function on `PrefixPart` objects. Since we now move the `search_ancestor` function to be a method of `NodeOrLeaf`, and `PrefixPart` is not a subclass of `NodeOrLeaf`, we need to manually add the `search_ancestor` method to `PrefixPart`. Otherwise, I think making `PrefixPart` a subclass of `NodeOrLeaf` may cause more problems.

parso/tree.py
gousaiyang: The `None` as default is simply borrowed from [`ast.dump`](https://docs.python.org/3/library/ast.html#ast.dump). Personally I always use `indent=4` for good readability, while `indent=None` is intended for a compact representation.

gousaiyang:
Also I've updated `parso.tree.search_ancestor` to contain the original actual implementation instead of calling `NodeOrLeaf.search_ancestor`. I'm not sure whether there are any users calling `parso.tree.search_ancestor` on non-`NodeOrLeaf` objects, but let's better be safe not to break things. I think a deprecated feature should continue working as-is (apart from possibly emitting a warning) until it is removed.

parso/python/tree.py
```diff
@@ -549,7 +548,15 @@ class Function(ClassOrFunc):
...
-        parameters.children[1:-1] = _create_params(parameters, parameters.children[1:-1])
+        parameters_children = parameters.children[1:-1]
+        input_has_param = False
+        for child in parameters_children:
+            if isinstance(child, Param):
+                input_has_param = True
+        # If input parameters list already has Param objects, keep it as is;
+        # otherwise, convert it to a list of Param objects.
+        if not input_has_param:
+            parameters.children[1:-1] = _create_params(parameters, parameters_children)
```
isidentical, line 559: Can we use something like this here instead of an expanded for loop with variable overwrites (same goes below)?
`not any(isinstance(child, Param) for child in parameters.children[1:-1])`

parso/tree.py
isidentical: `None` on `ast.dump` was just to preserve backwards compatibility since a lot of people used that function. Since we are just creating this function, it would be cool to have  `4` as the default.

isidentical (approved):
looks amazing!

parso/tree.py
davidhalter: Thanks!

parso/tree.py
gousaiyang: Yes, I see the `indent` option of `ast.dump` was just added in Python 3.9, and previously it only generates the single-line compact representation. Let's use `4` for our new function to pretty-print by default.

parso/python/tree.py
gousaiyang: Yes we can definitely do that. The old version of this PR has more logic under the `if` branch and cannot be simplified into an `any` call with comprehension. Later the logic was removed but I forgot to simplify the whole piece of code.

parso/tree.py
davidhalter: Great, I think it looks good now. Thanks for changing it! I hope the diff parser does not fall flat on its face, but we'll find out :)

PR #188, opened 2021-05-30
Fix line endings support at various locations

gousaiyang:
Some parts of the code do not correctly handle all kinds of line endings (LF, CRLF, CR), which needs to be fixed. Specifically, `test_normalizer_issues_files.py` fails on Windows if the repository was checked out with Git `core.autocrlf` option set to `true`.


davidhalter:
Great find. Thanks!

PR #194, opened 2021-08-05
Update expected exception line numbers for Python 3.10.0rc1

mgorny:
It seems that upstream has fixed line numbers in some of the expections
in Python 3.10.0rc1, so update the tests accordingly.  This means that
test_non_async_in_async() gets the correct line again,
and test_default_except_error_postition() no longer suffers from
the apparent off-by-one problem.

This doesn't fix tests entirely with Python 3.10 but it's a step
forward.

### DIFF
--- a/test/test_python_errors.py
+++ b/test/test_python_errors.py
@@ -57,10 +57,10 @@ def test_non_async_in_async():
         error, = errors
         actual = error.message
     assert actual in wanted
-    if sys.version_info[:2] < (3, 8):
+    if sys.version_info[:2] not in ((3, 8), (3, 9)):
         assert line_nr == error.start_pos[0]
     else:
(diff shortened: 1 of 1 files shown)


mgorny:
NB: I've skipped AUTHORS.txt for this tiny change but if you believe I should add myself anyway, please lemme know.

isidentical (approved):
Thanks for the patch, due to PEP 626 there have been various fixes overall about line numbers of particular expression/statements and probably these got fixed a long the way.

davidhalter:
There's a small linting issue. Can you fix that?

Otherwise thanks for the patch!

mgorny:
I'm sorry about that typo. Should be fixed now.

PR #199, opened 2021-09-14
Fixes `__slots__` definition in `NodeOrLeaf`

sobolevn:
This place: 
https://github.com/davidhalter/parso/blob/ae491cbf55853abf9c2cf3ab36ed1b9adc46384e/parso/tree.py#L385
uses `.parent` assignment on `NodeOrLeaf` type. While, `parent` was not in `__slots__` of `NodeOrLeaf`. But in `__slots__` of both its children: `Leaf` and `BaseNode`.

Identified by https://github.com/python/mypy/pull/10864

### DIFF
--- a/parso/tree.py
+++ b/parso/tree.py
@@ -26,7 +26,7 @@ class NodeOrLeaf:
     """
     The base class for nodes and leaves.
     """
-    __slots__ = ()
+    __slots__ = ('parent',)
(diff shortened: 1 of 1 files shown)


davidhalter:
Thanks for making the PR! This was probably not a real issue for anybody, but it's still a good fix.

PR #208, opened 2022-09-25
Use --no-implicit-optional for type checking

hauntsaninja:
This makes type checking PEP 484 compliant (as of 2018).
mypy will change its defaults soon.

See:
https://github.com/hauntsaninja/no_implicit_optional
https://github.com/python/mypy/issues/9091
https://github.com/python/mypy/pull/13401

### DIFF
--- a/parso/grammar.py
+++ b/parso/grammar.py
@@ -47,15 +47,15 @@ class Grammar(Generic[_NodeT]):
         self._hashed = hashlib.sha256(text.encode("utf-8")).hexdigest()
 
     def parse(self,
-              code: Union[str, bytes] = None,
+              code: Optional[Union[str, bytes]] = None,
               *,
               error_recovery=True,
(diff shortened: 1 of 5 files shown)


isidentical (approved):

davidhalter:
Why? I like implicit optionals.

hauntsaninja:
Okay, I'll close this in favour of https://github.com/davidhalter/parso/pull/209 then :-)

Is your question why PEP 484 was changed? In the early days of type checking in Python, type checkers played fast and loose with None, basically treating it the way Java treats null (anything can be None at any point in time and mypy won't complain about it). Once that was changed, this meant this was the only place in the type system where a variable isn't of the type its type annotation says it is, which is pretty surprising. It also contributed to the common confusion of what Optional means: many new typing users conflate it with whether or not an argument can be omitted (because it has a default value).

PR #209, opened 2022-09-27
Explicitly allow implicit optionals

hauntsaninja:
Keeps things working when you upgrade mypy versions. The other way of solving this problem is in #208


davidhalter:
Great, thanks a lot for looking at this again!

Issue #192, opened 2021-06-29
Multiple test/test_python_errors.py failures with Python 3.10


mgorny:
Full pytest output follows:

```
============================= test session starts ==============================
platform linux -- Python 3.10.0b3, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /tmp/parso, configfile: pytest.ini, testpaths: parso, test
plugins: case-1.5.3, xprocess-0.17.1, mock-3.6.1, expect-1.1.0, forked-1.3.0, asyncio-0.15.1, freezegun-0.4.2, timeout-1.4.2, datadir-1.3.1, metadata-1.8.0, snapshottest-0.5.1, rerunfailures-10.0, subtests-0.5.0, localserver-0.5.0, betamax-0.8.1, pyfakefs-4.5.0, httpbin-1.0.0, django-4.4.0, flaky-3.7.0, hypothesis-6.14.0, shutil-1.7.0, virtualenv-1.7.0, pkgcore-0.12.1, xdist-2.3.0, cov-2.12.1, Faker-8.8.2
collected 1351 items

parso/__init__.py .                                                      [  0%]
parso/tree.py .                                                          [  0%]
parso/python/tree.py .                                                   [  0%]
test/test_cache.py .........                                             [  0%]
test/test_diff_parser.py ............................................... [  4%]
...................................                                      [  6%]
test/test_dump_tree.py ..........                                        [  7%]
test/test_error_recovery.py .............                                [  8%]
test/test_file_python_errors.py .....                                    [  9%]
test/test_fstring.py ................................................... [ 12%]
..................                                                       [ 14%]
test/test_get_code.py .............                                      [ 15%]
test/test_grammar.py .                                                   [ 15%]
test/test_load_grammar.py ...........                                    [ 15%]
test/test_normalizer_issues_files.py .............................       [ 18%]
test/test_old_fast_parser.py ...............                             [ 19%]
test/test_param_splitting.py .......                                     [ 19%]
test/test_parser.py .................................................... [ 23%]
........................................................................ [ 28%]
..................                                                       [ 30%]
test/test_parser_tree.py ............................................... [ 33%]
................................                                         [ 36%]
test/test_pep8.py ...                                                    [ 36%]
test/test_pgen2.py ..................................................... [ 40%]
........................................................................ [ 45%]
........................................................................ [ 50%]
........................................................................ [ 56%]
..                                                                       [ 56%]
test/test_prefix.py ......................                               [ 58%]
[...]

PR #211, opened 2022-12-04
Fix unit tests in Python 3.10 (Closes: #192)

jspricke:


jspricke:
I've pushed a new version working on Python 3.9 (Debian stable), 3.10 and 3.11 (Debian unstable). Can you trigger the CI again? The initial patch was for Debian unstable where we don't have older Python versions.

davidhalter:
This time it looks much better. I feel like the linter should still be fixed, but 3.6 is probably something I have to look at.

davidhalter:
Is it allowed that I can push changes to this branch? Because I would probably fix the Python 3.6 error this way.

davidhalter:
I just realized that 3.6 has been end of life for a year. Can you just remove it from the CI? I feel like that's probably the best way forward.

jspricke:
I've updated the Python versions, can you trigger the CI again?
Also feel free to push to this branch.

davidhalter:
Unfortunately still some errors. I have tried to change the rules when the tests are run. So I feel like the tests will be run now when anybody not new to GitHub opens a PR. So I feel like you should now be able to see the results quicker.

jspricke:
Thanks for changing the CI restrictions. I assume the reduced coverage is ok? (all other tests pass now)

PR #218, opened 2023-03-12
fix out-by-one in get_leaf_for_position

### LINKED ISSUE #217: get_leaf_for_position returns incorrect element
The binary search in `get_leaf_for_position()` may return a leaf whose end position is the position being queried for.

But AFAICS the end position is intended to be exclusive (which is what I would expect).

If I have not misunderstood the fix is straightforward and a pull request will follow.

(https://github.com/pappasam/jedi-language-server/commit/1a75efb491599e6ba69d7c08b637f881586dddb5 is what brought me here: that's a hokey workaround that seems to kludge its way around the issue, mostly)

dimbleby:
Fixes #217

I've taken the liberty of rewriting the binary search so that it's not recursive.  To my tastes that's much easier to follow but I guess if you really liked it the way it was I can switch it back!

### DIFF
--- a/parso/tree.py
+++ b/parso/tree.py
@@ -414,28 +414,29 @@ class BaseNode(NodeOrLeaf):
             on whitespace or comments before a leaf
         :return: :py:class:`parso.tree.Leaf` at ``position``, or ``None``
         """
-        def binary_search(lower, upper):
-            if lower == upper:
-                element = self.children[lower]
-                if not include_prefixes and position < element.start_pos:
-                    # We're on a prefix.
-                    return None
-                # In case we have prefixes, a leaf always matches
-                try:
-                    return element.get_leaf_for_position(position, include_prefixes)
-                except AttributeError:
-                    return element
-
-            index = int((lower + upper) / 2)
-            element = self.children[index]
-            if position <= element.end_pos:
-                return binary_search(lower, index)
-            else:
-                return binary_search(index + 1, upper)
-
-        if not ((1, 0) <= position <= self.children[-1].end_pos):
+        if not ((1, 0) <= position < self.children[-1].end_pos):
             raise ValueError('Please provide a position that exists within this node.')
-        return binary_search(0, len(self.children) - 1)
+
+        # Binary search for matching child.
+        lower = 0
+        upper = len(self.children) - 1
+        while lower < upper:
+            middle = int((lower + upper) / 2)
+            element = self.children[middle]
+            if position < element.end_pos:
+                upper = middle
+            else:
+                lower = middle + 1
+
+        element = self.children[lower]
+        if not include_prefixes and position < element.start_pos:
+            # We're on a prefix.
+            return None
+        # In case we have prefixes, a leaf always matches
+        try:
+            return element.get_leaf_for_position(position, include_prefixes)
+        except AttributeError:
+            return element
 
     def get_first_leaf(self):
         return self.children[0].get_first_leaf()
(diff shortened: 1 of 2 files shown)


dimbleby:
in fact let's not write our own binary search, there's one in the standard library.

Just need to make sure we use the right one out of `bisect_left()` vs `bisect() / bisect_right()`

parso/tree.py
```diff
@@ -414,28 +415,19 @@ def get_leaf_for_position(self, position, include_prefixes=False):
...
+
+        index = bisect(self.children, position, key=lambda child: child.end_pos)
+        element = self.children[index]
+        if not include_prefixes and position < element.start_pos:
+            # We're on a prefix.
+            return None
+        # In case we have prefixes, a leaf always matches
+        try:
+            return element.get_leaf_for_position(position, include_prefixes)
+        except AttributeError:
```
davidhalter, line 429: Sorry, but I don't want this. Why is this happening? Catching AttributeError in such a general fashion is something I never accept.

I'm fine with a non-recursive solution. Using bisect looks like a good idea. However this is something I really really want to avoid.

dimbleby:
ah no, the `key` on the standard library's bisect is new.  So do write our own after all, sorry for the noise

davidhalter (commented):
Thanks for the patch! I will have to test with Jedi if all the tests still pass, but in general I feel like this is a good direction.

parso/tree.py
dimbleby: this is just the existing code, moved

davidhalter: Oh god. Doesn't make it better, but thanks for pointing it out.

dimbleby: I guess you could instead ask whether the leaf is an instance of `BaseNode`, or even whether it has the `get_leaf_for_position` attribute.

The comment here seems confusing, I think what it really means is something like this:
```python
        # If we've found another node, recurse.
        if isinstance(element, BaseNode):
            return element.get_leaf_for_position(position, include_prefixes)

        return element
```

can make that change if you want - though I'm beginning to sense possible mission creep.  Let me know

dimbleby:
well this causes absolute carnage in jedi unit tests, I might need some help understanding why that is and whether this is salvageable...

davidhalter:
Is there a way you can tell me what you changed without the refactoring? I have a hard time understanding that. I don't really understand the original issue either.

You don't have to revert all the changes, but maybe just post a few lines that you would change in the old code?

parso/tree.py
davidhalter: If it's really that easy I usually write it like this:

```
try:
    method = element.get_leaf_for_position
expect AttributeError:
    ...
else:
    return method(position, include_prefixes)
```

But let's first figure out the other problems.

dimbleby:
without the refactoring the change is just to the condition ` if position < element.end_pos`, which originally was `<=`

the unit test is intended to show the problem: without the fix the element that is found is not the one at the given position.

davidhalter:
This is intentional. It is unclear if the position is on the `\n` or on the next node. So this is kind of intentional, I guess. So I feel like we need to investigate why this has happened.

Jedi usually does this by first asking for `get_name_of_position` and only if no name is found it asks for `get_leaf_for_position`. So I guess the mistake is that we don't special case `\n` there. In that case (and if the next token is starting at the same point), we should be checking it there. So I'm inclined to close, because this should be done in Jedi. Do you agree? Do you want to work it out there?

dimbleby:
I was coming to a similar conclusion: or at any rate that making a change along those lines was less likely to break the world

Round about https://github.com/davidhalter/jedi/blob/b814ca2951b29be2a103395039240dbe6c5c154d/jedi/api/__init__.py#L371, something like this, perhaps?

```patch
+
+        if leaf is not None and leaf.end_pos == (line, column) and leaf.type == 'newline':
+            next_ = leaf.get_next_leaf()
+            if next_ is not None and next_.start_pos == leaf.end_pos:
+                leaf = next_
+
```

dimbleby:
https://github.com/davidhalter/jedi/pull/1923

PR #78, opened 2019-07-04
ENH: update grammar for py39

tacaswell:

PR #220, opened 2023-06-18
ENH: add grammar file from py313

tacaswell:
Following https://github.com/davidhalter/parso/pull/78 copied the py312 grammar file (which looks like it also a copy of the py311 file).


davidhalter:
Thanks!

PR #230, opened 2025-03-10
Update setup.py classifiers

Erotemic:
Current classifiers in setup.py are missing supported versions of Python. This PR adds them.

It looks like the 314 grammar is there, but I didn't see any tests for it in CI, so leaving that out for now.


davidhalter:
Thanks!

PR #233, opened 2025-08-21
Fix CI

A5rocks:
It seems 3.7 is breaking CI... it's also ~2 years past EOL. It's probably OK to remove it from CI.

I then checked other files for mentions of 3.7 and found 2 places that weren't up to date. Hopefully they're fine to update. I didn't remove any code adding support for testing 3.7 because... well... I don't know, that seems out of scope.


davidhalter:
Good question. I think it would make sense to drop old grammars as well, people will just run into problems there.

Issue #232, opened 2025-05-09
Fallback to previous Python grammar versions if there is no grammar for the current one


jsbueno:
Python has usually two ongoing pre-release versions - and parso is used by the autocomplete from different shells.

Currently, the only thing that makes using a shell like ipython subpar with a pre-release version (either 3.14 beta or 3.15 alpha as of this report) is parso's exception on trying to load grammars for those versions. 
(an otherwise very pleasant REPL session for testing/showing new features becomes overwhelmed by traceback errors due to the grammar file not being found).

I have found a personal workaround of simply copying the latest grammar around - but that (1) takes a bit of time, even after you already know about it; (2) have to be done again for each new virtualenv 

What if instead of erroring out for newer/not found versions, Parso would show a warning the first time, and them simply load the latest available grammar file? 

-------------
(offending line: https://github.com/davidhalter/parso/blob/9328cffce392b4c3e907db4bbfad06ca1111b3ee/parso/grammar.py#L264 )

davidhalter:
Yeah core devs are probably affected by this the most. Usually somebody copies the latest grammar and creates a PR. But I agree that this should probably load the latest available grammar (especially since have not been fully compatible with the latest grammar anyway).

PR appreciated, I'm happy to merge. Even just hardcoding it (where we raise NotImplementedError now) would probably be better than the current behavior. I'm not sure we need to do some big magic there.

PR #234, opened 2025-08-21
Load newest grammar in face of a future grammar

A5rocks:
This fixes https://github.com/davidhalter/parso/issues/232.

This probably needs https://github.com/davidhalter/parso/pull/233 and I'll rebase once that's merged. Hopefully CI is fine with my changes!


parso/grammar.py
```diff
@@ -239,7 +240,15 @@ def load_grammar(*, version: str = None, path: str = None):
...
-    version_info = parse_version_string(version)
+    # NOTE: this (3, 14) should be updated to the latest version parso supports.
+    #       (if this doesn't happen, users will get older syntaxes and spurious warnings)
+    passed_version_info = parse_version_string(version)
+    version_info = min(passed_version_info, PythonVersionInfo(3, 14))
+
+    if passed_version_info != version_info:
+        warnings.warn('parso does not support %s.%s yet.' % (
+            passed_version_info.major, passed_version_info.minor
+        ))
```
davidhalter, line 251: That's a really clever idea and I like the general direction, but I feel like we should not raise a warning. Maybe comment this so I can re-enable it if I ever properly start using 3.10+ Python syntax.

I currently have different priorities (https://zubanls.com/ where a 3.13 Python parser is already included and I think it will soon be open sourced), so I think this warning will just be annoying to people and I think we could probably also delete *most* grammar files, because they are currently just copies of 3.9 IIRC. But you don't have to do that, I just think that simply copying the previous parso version over and over again is not very useful... It used to be that these files were actually different.

Or do you think the warnings makes a lot of sense?

In any way, please keep the:

```
# Disallow warnings
filterwarnings = error
```

I like that regardless.

PR #236, opened 2026-02-18
Fix parsing of raw f-strings with backslash before brace

### LINKED ISSUE #207: KeyError: ReservedString(})
## Minimal example

```python
import parso

parso.parse('stream.write(b_(rf"\{ord_(c):0>3o}"))', error_recovery=False)
```

## Traceback

```
Traceback (most recent call last):
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/parser.py", line 180, in _add_token
    plan = stack[-1].dfa.transitions[transition]
KeyError: ReservedString(})

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/__init__.py", line 58, in parse
    return grammar.parse(code, **kwargs)
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/grammar.py", line 156, in parse
    root_node = p.parse(tokens=tokens)
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/python/parser.py", line 79, in parse
    return super().parse(tokens)
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/parser.py", line 129, in parse
    self._add_token(token)
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/parser.py", line 186, in _add_token
    self.error_recovery(token)
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/python/parser.py", line 140, in error_recovery
    return super().error_recovery(token)
  File "/home/moose/.pyenv/versions/3.10.2/lib/python3.10/site-packages/parso/parser.py", line 152, in error_recovery
    raise ParserSyntaxError('SyntaxError: invalid syntax', error_leaf)
parso.parser.ParserSyntaxError: ('SyntaxError: invalid syntax', <ErrorLeaf: PythonTokenTypes.OP:'}', (1, 33)>)
```

## Environment

```
$ python --version     
Python 3.10.2

>>> parso.__version__
'0.8.3'
```

## Follow-up issues

This caused mutmut to fail: https://github.com/boxed/mutmut/issues/252

worksbyfriday:
## Summary

Fixes #207. Raw f-strings like `rf"\{x}"` fail to parse because the tokenizer treats `\{` as a backslash escape sequence, consuming the `{` and preventing it from opening an f-string expression.

### Root Cause

The f-string content regex (`fstring_string_single_line`) includes `\\[^\r\nN]` which matches a backslash followed by any non-newline character. This is correct for non-raw f-strings where `\n`, `\t`, etc. are escape sequences. But in raw f-strings, backslash is a literal character -- `\{` should be tokenized as literal `\` followed by expression-opening `{`.

Parso used the same regex for both raw and non-raw f-strings because `FStringNode` only tracked the quote character, not the prefix.

### Fix

- Add `is_raw` flag to `FStringNode`
- Store the raw flag in `fstring_pattern_map` alongside the quote
- Add separate raw-aware regex patterns (`fstring_string_single_line_raw`, `fstring_string_multi_line_raw`) that treat backslash as an ordinary character
- Select the appropriate regex in `_find_fstring_string` based on `tos.is_raw`

### Tests

Added 8 test cases for raw f-strings with all prefix case variations (`rf`, `Rf`, `rF`, `fR`, `Fr`) and both single and triple-quoted forms. All 1812 tests pass.

### DIFF
--- a/parso/python/tokenize.py
+++ b/parso/python/tokenize.py
@@ -119,6 +119,10 @@ fstring_string_single_line = _compile(
 fstring_string_multi_line = _compile(
     r'(?:\{\{|\}\}|\\N\{' + unicode_character_name + r'\}|\\[^N]|[^{}\\])+'
 )
+# In raw f-strings, backslash is a literal character and does NOT escape
+# the following character.  So \{ means literal \ followed by expression {.
+fstring_string_single_line_raw = _compile(r'(?:\{\{|\}\}|[^{}\r\n])+')
+fstring_string_multi_line_raw = _compile(r'(?:\{\{|\}\}|[^{}])+')
 fstring_format_spec_single_line = _compile(r'(?:\\(?:\r\n?|\n)|[^{}\r\n])+')
 fstring_format_spec_multi_line = _compile(r'[^{}]+')
 
@@ -219,7 +223,7 @@ def _create_token_collection(version_info):
 
     for t in fstring_prefixes:
         for quote in all_quotes:
-            fstring_pattern_map[t + quote] = quote
+            fstring_pattern_map[t + quote] = (quote, 'r' in t.lower())
 
     ALWAYS_BREAK_TOKENS = (';', 'import', 'class', 'def', 'try', 'except',
                            'finally', 'while', 'with', 'return', 'continue',
@@ -253,8 +257,9 @@ class PythonToken(Token):
 
 
 class FStringNode:
-    def __init__(self, quote):
+    def __init__(self, quote, is_raw=False):
         self.quote = quote
+        self.is_raw = is_raw
         self.parentheses_count = 0
         self.previous_lines = ''
         self.last_string_start_pos: Any = None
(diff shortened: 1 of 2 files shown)


davidhalter:
The linter does not pass. The coveralls error is obviously fine. Something is wrong with the service.

Since this is likely generated/assisted with AI, please make sure to think about this problem well enough so I don't have to. It really really wastes my time and makes me mad if people don't. Also I would personally recommend to not use AI for contributions, they are in my experience just sub-par and you don't learn anything yourself doing it that way.

worksbyfriday:
Understood and respected. Thank you for explaining your position clearly. Closing this.

PR #237, opened 2026-03-31
PEP 695 type parameter syntax for Python 3.12+

### LINKED ISSUE #221: parso can't parse PEP 695: Type Parameter Syntax
Hi, first of all thank you for creating awesome library.
I'm useing parso in my library.

Python3.12 has big change at grammer.
https://docs.python.org/3/whatsnew/3.12.html#whatsnew312-pep695

parso does not support PEP 695: Type Parameter Syntax.

```python
import parso

code = """def func[T](arg1: T, arg2: T) -> T:
    return arg1 + arg2"""

module = parso.parse(code, version='3.12')
expr = module.children[0]
print(expr) 
print(expr.get_code()) 
```
output
```
<PythonErrorNode: def func@1,0>
def func
```

This example is def, but classes may be similar.

It would be very helpful if you could support the new grammar.
Thank you.

darki73:
Closes #221

I ran into this while building a code intelligence tool that uses jedi for cross-module resolution. My ORM uses `class QueryBuilder[T: "Model"](Mixin1, Mixin2)` and jedi couldn't resolve any of the inherited methods. Traced it back to parso parsing the class as an error node.

Two things were broken:

1. The 3.12/3.13/3.14 grammars don't have `type_params` rules, so `class Foo[T]` and `def bar[T]()` parse as error nodes.

2. `Class.get_super_arglist()` hardcodes `children[2]` and `children[3]` to find the arglist. With type params present, `(` shifts to a later position and the method returns None. This means jedi can't find base classes at all, so the entire MRO is invisible.

Fix: added `type_params`, `type_param`, `type_param_bound` rules to the grammars. Added `type_param_default` for 3.13+ (PEP 696). Replaced the hardcoded index in `get_super_arglist` with a search for `(`.

Worth noting - in the jedi issue (davidhalter/jedi#2025) there was concern that jedi would need major changes to support this. Turns out it doesn't. Once parso parses the class correctly and `get_super_arglist` returns the right node, jedi's existing MRO resolution handles everything. Completions, goto, inference on inherited methods - all work without touching jedi at all. The `get_super_arglist` fix was the missing piece.

The `type` soft keyword statement is not included here since that needs tokenizer changes.

### DIFF
--- a/parso/python/tree.py
+++ b/parso/python/tree.py
@@ -479,13 +479,13 @@ class Class(ClassOrFunc):
         Returns the `arglist` node that defines the super classes. It returns
         None if there are no arguments.
         """
-        if self.children[2] != '(':  # Has no parentheses
-            return None
-        else:
-            if self.children[3] == ')':  # Empty parentheses
-                return None
-            else:
-                return self.children[3]
+        for i, child in enumerate(self.children):
+            if child == '(':
+                next_child = self.children[i + 1]
+                if next_child == ')':
+                    return None
+                return next_child
+        return None
 
 
 def _create_params(parent, argslist_list):
@@ -552,15 +552,21 @@ class Function(ClassOrFunc):
 
     def __init__(self, children):
         super().__init__(children)
-        parameters = self.children[2]  # After `def foo`
+        parameters = self._find_parameters()
         parameters_children = parameters.children[1:-1]
-        # If input parameters list already has Param objects, keep it as is;
-        # otherwise, convert it to a list of Param objects.
         if not any(isinstance(child, Param) for child in parameters_children):
-            parameters.children[1:-1] = _create_params(parameters, parameters_children)
+            parameters.children[1:-1] = _create_params(
+                parameters, parameters_children
+            )
+
+    def _find_parameters(self):
+        for child in self.children:
+            if child.type == 'parameters':
+                return child
+        raise Exception("A function should always have parameters")
 
     def _get_param_nodes(self):
-        return self.children[2].children
+        return self._find_parameters().children
 
     def get_params(self):
         """
@@ -633,13 +639,10 @@ class Function(ClassOrFunc):
         """
         Returns the test node after `->` or `None` if there is no annotation.
         """
-        try:
-            if self.children[3] == "->":
-                return self.children[4]
-            assert self.children[3] == ":"
-            return None
-        except IndexError:
-            return None
+        for i, child in enumerate(self.children):
+            if child == '->':
+                return self.children[i + 1]
+        return None
 
 
 class Lambda(Function):
--- a/conftest.py
+++ b/conftest.py
@@ -145,3 +145,15 @@ def works_ge_py38(each_version):
 def works_ge_py39(each_version):
     version_info = parse_version_string(each_version)
     return Checker(each_version, version_info >= (3, 9))
+
+
+@pytest.fixture
+def works_ge_py312(each_version):
+    version_info = parse_version_string(each_version)
+    return Checker(each_version, version_info >= (3, 12))
+
+
+@pytest.fixture
+def works_ge_py313(each_version):
+    version_info = parse_version_string(each_version)
+    return Checker(each_version, version_info >= (3, 13))
--- a/parso/python/grammar312.txt
+++ b/parso/python/grammar312.txt
@@ -17,7 +17,11 @@ decorators: decorator+
 decorated: decorators (classdef | funcdef | async_funcdef)
 
 async_funcdef: 'async' funcdef
-funcdef: 'def' NAME parameters ['->' test] ':' suite
+funcdef: 'def' NAME [type_params] parameters ['->' test] ':' suite
+
+type_params: '[' type_param (',' type_param)* [','] ']'
+type_param: NAME [type_param_bound] | '*' NAME | '**' NAME
+type_param_bound: ':' test
 
 parameters: '(' [typedargslist] ')'
 typedargslist: (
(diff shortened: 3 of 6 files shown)


davidhalter:
Did you run the tests in Jedi with this changed parso version? And did they all pass?

Thanks a lot for trying to work on this!

darki73:
I am using jedi 0.19.2 as a direct dependency in my project, before submitting PR for parso, I "patched" the loading mechanism of parso in my projects code because it basically allowed me to submit the version I wanted to release without the need to wait for the PR to be merged.

So far, on 62 cases without the patch, this patch provides 62/62 mappings (i am, like parso, also run my app on itself for checks) - https://github.com/darki73/sylvan/blob/899de480992eb79f1ba8f3c71ac8552b68e647fc/src/sylvan/analysis/structure/jedi_setup.py

This is pretty much the same "fix", just monkey patched.

I have not though ran any tests in jedi, i can try to do it later today / tomorrow when I have time to do so.

Thank you for taking time to review the PR!

darki73:
@davidhalter 

Ran the tests against the main of jedi, found some failing tests that required more changes on the parso side (which are already added to this PR as of now. `Function` alongside with `Class` changes initially proposed).

As of now, 3743 tests in jedi have passed (had some issues with windows, but turns out just flaky tests on this specific platform).

parso/python/tree.py
```diff
@@ -479,13 +479,15 @@ def get_super_arglist(self):
...
         """
-        if self.children[2] != '(':  # Has no parentheses
-            return None
-        else:
-            if self.children[3] == ')':  # Empty parentheses
-                return None
-            else:
-                return self.children[3]
+        for i, child in enumerate(self.children):
+            if hasattr(child, 'value') and child.value == '(':
```
davidhalter: Can you use

```
try:
    value = child.value
except AttributeError:
    continue
if value == "(":
    ...
```

here and in the places below? `hasattr` should not be used like that because it uses multiple lookups and is harder to refactor.

darki73: Actually, looked at the existing patterns in this file, we might even go with the direct string comparison.

(had to edit the next line, was confusing as hell to "unprepared" person)
Basically, the original code used the `!=` operator to exit early, and since we are in the loop now, we are checking with `==` for a `(` instead of its absence.
So just like before, instead of `self.children[2] != '('` we can just do `child == '('`.

I will push the updated version for both places, but ultimately, it is for you to decide as you are the author and if you prefer the try/except - just let me know and i will update it.

With this, 1984 parso and 3743 jedi tests pass no issue.

parso/python/tree.py
```diff
@@ -633,13 +645,11 @@ def annotation(self):
...
-        try:
-            if self.children[3] == "->":
-                return self.children[4]
-            assert self.children[3] == ":"
-            return None
-        except IndexError:
-            return None
+        for i, child in enumerate(self.children):
+            if child == '->':
+                if i + 1 < len(self.children):
```
davidhalter: I think this should always be true, since the grammar file specifies this. I think this line can just be removed.

parso/python/tree.py
```diff
@@ -552,15 +554,25 @@ class Function(ClassOrFunc):
...
     def __init__(self, children):
         super().__init__(children)
-        parameters = self.children[2]  # After `def foo`
-        parameters_children = parameters.children[1:-1]
-        # If input parameters list already has Param objects, keep it as is;
-        # otherwise, convert it to a list of Param objects.
-        if not any(isinstance(child, Param) for child in parameters_children):
-            parameters.children[1:-1] = _create_params(parameters, parameters_children)
+        parameters = self._find_parameters()
+        if parameters is not None:
```
davidhalter: This should also never be reachable and can be removed. A function always has parameters, see grammar file

parso/python/tree.py
```diff
@@ -552,15 +554,25 @@ class Function(ClassOrFunc):
...
+            if not any(isinstance(child, Param) for child in parameters_children):
+                parameters.children[1:-1] = _create_params(
+                    parameters, parameters_children
+                )
+
+    def _find_parameters(self):
+        for child in self.children:
+            if child.type == 'parameters':
+                return child
+        return None
```
davidhalter: This should just `raise Exception("A function should always has parameters")`.

parso/python/tree.py
```diff
@@ -479,13 +479,15 @@ def get_super_arglist(self):
...
-        if self.children[2] != '(':  # Has no parentheses
-            return None
-        else:
-            if self.children[3] == ')':  # Empty parentheses
-                return None
-            else:
-                return self.children[3]
+        for i, child in enumerate(self.children):
+            if child == '(':
+                if i + 1 >= len(self.children):
```
davidhalter: Same here, I guess this should always be true.

davidhalter (commented):
Sorry for all the comments, I like the direction and probably prefer that to the AttributeError stuff, I just think that while it's hard to understand the code, it gets much easier if you know/look at the grammar. Does that help with understanding it for you? I don't think it's my goal that everybody understands it without looking at the grammar files.

darki73:
No problem!
I just went a bit defensive on the python side instead of just "trusting" the grammar parser.
But you are right, no need for those check specifically because grammar itself ensures that there is just no possible way for the parameters not to be there.

davidhalter:
Thanks a lot again for working on this! I think it's especially meaningful, because it fixes things without causing too much work in other places.

And feel free to checkout Zuban, I'm currently in the process of bringing some Jedi goodness to it and there should probably not be a lot of reasons anymore in a few months to use Jedi instead of Zuban.

delfick:
@darki73 @davidhalter excuse me for the noise, but thank you so much for fixing this!!!!!!

PR #245, opened 2026-09-22
docs: distinguish tree positions from prefixes

Likio3000:
The parser-tree API docstrings describe `start_pos` and `end_pos` as positions of the prefix, although the implementations report the node or token boundaries. They also describe `include_prefix` as removing the prefix without explaining that this happens when it is false.

Distinguish node/leaf positions from prefix positions, correct the existing `+` example's terminology, and explain that `get_code(include_prefix=False)` removes only the leading prefix while retaining prefixes inside a node. Add a documentation contributor entry as requested by CONTRIBUTING.

Validation: statically reviewed the `Leaf` and `BaseNode` implementations and existing tree/code-preservation assertions. Python ASTs are identical after removing docstrings; exact original blobs and the diff were checked. No parso import, test suite, Sphinx build or project script was run. This changes four API docstrings and AUTHORS only.

Prepared with AI assistance (OpenAI Codex), with source and diff review.


davidhalter:
Thanks

